PR #179, opened 2014-11-01
plone.api.folder added 

### LINKED ISSUE #177: Suggestion: method for retrieving folder contents
plone.api currently does have a canonical API method for retrieving the contents of a folderish object.
Something like

  contents = plone.api.folder.list_contents(container, searchFilters, brains=False)

The semantic is obvious since it it modeled after the getFolderContents() method as we have it already in Plone. 'searchFilters' is a dict of query parameters for the portal_catalog. By default the method would return objects instead of brains in order to avoid brains from the n00b users. Advanced users could use brains=True for large folders if needed and wanted.

zopyx:
closes #177

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -14,7 +14,7 @@ Changelog
   ``api.user.revoke_roles``
   [adamcheasley]
 
-- added ``api.folder.list_objects()`` and ``api.folder.list_brains()``
+- Added ``api.folder.list_objects()`` and ``api.folder.list_brains()``
   [ajung]
 
 1.2.1 (2014-06-24)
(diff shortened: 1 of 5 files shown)


jensens:
I'd like to see this one merged. Some minor polishing needed. Maybe its easier if we first take this over as a branch in ``plone/plone.api`` in order to be able to fix the last minor style issues?

gforcada:
I would very much like the commits to be squashed together and have more verbose commit messages (as the style guide recommends anyway).

adamcheasley:
I think taking a new branch here is going to be the easiest thing. Possibly even close this pull request and open a new one once the styling issues are fixed.

gforcada:
So @zupo what should be done here? #181 is still open, should the release team have a last saying on it?

gforcada:
I just pushed https://github.com/plone/plone.api/tree/zopyx-plone.api.folder with only 4 commits one for each logic set of changes. We can continue the discussion on #205

PR #199, opened 2014-11-25
Working with images

pigeonflight:
An example of using the plone.api to create an image.


pbauer:
@pigeonflight This could be a part of the docs but not for plone.api. Also ``setImage`` is Archetypes only. In Dexterity you have to set contruct a instance of plone.namedfile.file.NamedBlobImage and set the image-attribute to it:
```python
from plone.namedfile.file import NamedBlobImage
image.image = NamedBlobImage(data=example_payload, filename='some_image.png')
```

Issue #181, opened 2014-11-01
Define process for including new modules to plone.api


zupo:
Recently there were ideas to add additional plone.api modules, such as api.folder and api.portlet. We need to define a process on how we accept new modules.

During the PloneConf 2014, @gforcada suggested we mimic how GTK does this and apply it to our use-case. After some discussion, this was what we came up with:
- for each new module, we will create a new package in the "experimenta.api.*" namespace
- there will be releases of these packages so people will be able to use them in the wild until we feel they are stable enough
- when we feel that an experimental package stabilized, it will be merged into plone.api -> at this point code using the "experimental.api._" can be rewritten to use "api._"

Example integration code with "portlets" module:

``` python
import experimental.api
from plone import api

api.content.create()
experimental.api.portlet.create()
```

jaroel:
What needs to be done, and who is going to do it?

PR #205, opened 2015-01-26
Zopyx plone.api.folder

gforcada:
Continuing #179 here. We need first to agree on #181 and then we can move this one accordingly.

Note that I just merged commits together so that the commits now do make sense, I did not change a single line of code (or this was the idea).

It's also rebased on top of master.

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -18,6 +18,10 @@ Changelog
   it doesn't set inherited roles locally.
   [cekk]
 
+- Added ``api.folder.list_objects()`` and ``api.folder.list_brains()``
+  [ajung]
+
+
 1.3.2 (2014-11-17)
 ------------------
 
--- a/docs/folder.rst
+++ b/docs/folder.rst
@@ -1,0 +1,62 @@
+.. admonition:: GitHub-only
+
+    WARNING: If you are reading this on GitHub, DON'T! Read the documentation
+    at `api.plone.org <http://api.plone.org/content.html>`_
+    so you have working references and proper formatting.
+
+
+.. module:: plone
+
+.. _chapter_folder:
+
+Content
+=======
+
+.. _list_objects_example:
+
+List folder content
+-------------------
+
+In order to retrieve the contents of a Plone folder or a folderish object you
+use the ``plone.api.folder.list_objects()`` method. The following example lists
+all Link objects with the ``some_folder`` folder object sorted by their title.
+You can query by all indexes that are available in portal_catalog and sort by
+all configured portal_catalog metadata. 
+
+.. code-block:: python
+
+    from plone import api
+    portal = api.portal.get()
+
+    api.content.create(container=portal, type='Document', id='document', title='Document')
+    api.content.create(container=portal, type='Link', id='link', title='Link')
+
+    objs = api.folder.list_objects(
+        container=portal,
+        content_filter={'portal_type': 'Link'},
+        sort_on='sortable_title')
+
+    assert objs[0].getId() == 'link'
+    assert objs[0].Title() == 'Link'
+
+    objs = api.folder.list_objects(
+        container=portal,
+        content_filter={'portal_type': 'Document'},
+        sort_on='sortable_title')
+
+    assert objs[0].getId() == 'document'
+    assert objs[0].Title() == 'Document'
+
+If your application does not need the full content objects and can work
+with Zcatalog brains then you can use the sister method 
+``plone.api.folder.list_brains()``:
+
+.. code-block:: python
+
+    from plone import api
+    brains = api.folder.list_brains(
+        container=portal,
+        content_filter={'portal_type': 'Link'})
+
+    assert brains[0].getId == 'link', objs[0].getId
+    assert brains[0].Title == 'Link', objs[0].Title
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -38,6 +38,7 @@ Narrative documentation
     about.rst
     portal.rst
     content.rst
+    folder.rst
     user.rst
     group.rst
     env.rst
@@ -52,6 +53,7 @@ Complete API and advanced usage
     api/index.rst
     api/portal.rst
     api/content.rst
+    api/folder.rst
     api/user.rst
     api/group.rst
     api/env.rst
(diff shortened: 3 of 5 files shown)


zupo:
Reading the comments, the #181 is agreed upon. It only needs a page in documentation with some narrative text and then what we said in #181.

jaroel:
We don't need two methods. plone.app.contentlisting provided a brain like interface, with a fallback to the actual object.
This is more in line with keeping plone.api simple and within 80% of usecases.
This might be of better use in plone.api.content.list_brains(context) and plone.api.content.list_objects(context), or plone.api.content.list(context)

Please discuss :)

zupo:
Hmm from the first look, I like the api.content.list() approach! No sure if it's better to have brains boolean parameter or different methods would be better though ...

adamcheasley:
I personally don't really see the need for listing as brains. Is the thinking here that if you had hundreds of objects, you might want to use the catalog instead, to avoid waking up objects?

zupo:
Ah true, we have the catalog. In this light, api.content.list(context) makes the most sense to me.

jaroel:
Please don't default to actual objects. Using just brains is fine, otherwise use .objectValues().

zupo:
Why not objects? Listing would return just first-level children of a given context, that shouldn't be bad in most cases?

jaroel:
Well, this api will be used in high traffic and high content sites too.
Preferably the main api we offer should not fail when we're sizing up.
In low content settings the overhead of brains is low, but in high content settings the objects is high.

jaroel:
Lets just do content.list(context) which returns brains and content.list_objects(context) that returns objects.

zupo:
What if we just document how to use plone.app.contentlisting instead? And not recreate (parts of) it ...

jaroel:
Sounds ok to me.

pbauer:
:+1: for encouraging the use of plone.app.contentlisting

jensens:
i think the idea is to have an api at one place? why not move the listing parts of p.a.contentlisting here and make it a consumer of plone.api?

jaroel:
I'd go the other way around. Use plone.api as a wrapper.

tisto:
Using plone.api as a wrapper is the same approach that I took for looking up the control panel settings. Idealy plone.api should just delegate to the original package, if that package has a sane API.

jensens:
In Bristol we had the idea to define packages which are consumers of plone.api and those that are producers for plone.api. 
We considered plone.app packages as possible consumers of plone.api. 

But since our dependency tree (well, graph or forrest would be better naming) is that horrible its at the moment impossible to define those right.

When we add features to plone.api we must think about this.

After reviewing the code of this new feature/PR I _don't_ think we should depend with plone.api on p.a.cl. Its pretty simple and basic functionality. To me it makes no sense at all to add a dependency here. We have enough already.

Other way around it makes eventually more sense, or we keep them separate at all? Keep things simple!

zupo:
Ideal world how I see it:
* plone.app.contentlisting has a sane basic UI, that does not require several imports and using multi adapters
* plone.api only contains documentation on basic usage of plone.app.contentlisting for listing content, and links to more advanced usage info on plone.app.contentlisting documentation

tisto:
@zupo +1

jensens:
Thinking more about it plone.api.folder adresses a python-developer centric problem, while p.a.cl adresses a front-end/template/UI centric problem (adding convinience). We may want to add a hint into the narrative documentation that there is actually p.a.cl available to address this problems.

jensens:
hint: plone.app.contentlisting is also planned to be merged to Products.CMFPlone (should be a pretty easy task) see https://dev.plone.org/ticket/13283

adamcheasley:
@jaroel Just to pick up on your comment regarding "high traffic and high content"; In our experience, when dealing with a site/application like that, the catalog pretty much falls over (when cataloged objects get into the hundreds of thousands). For this reason, we almost always use objectValues when we know where the objects are:
http://www.slideshare.net/hammertoe/mistakes-made-and-lessons-learnt-scaling-plone-postlaunch

See slide 19.

jensens:
indeed, and with dexterity things are changing anyway. i made the experience if the objects are small (i.e. basic folders, links, custom types, ...) its sometimes less expensive to iterate over objects than to search and wake up metadata.

jaroel:
Ack. How about filtering objects as proposed? We're dependent on the catalog's indexes for that, right?

jensens:
same with sorting. Btw, is metadata loaded with the brain or on first access of an value? If latter a getObject could be cheaper than loading metadata (vice versa we should fix the brain).

jaroel:
Looks like this should be catalog based (filtering and sorting). After that it doesn't really matter.
We can always get objects from brains, but preventing a object load after returning objects isn't possible.
I suggest using brains for content.list, wrapped in contentlisting which gets the object automatically.

jaroel:
Shouldn't we just use p.a.contentlisting's folderListing view? https://github.com/plone/plone.app.contentlisting/blob/master/plone/app/contentlisting/configure.zcml#L45

```python
def list(context, request=None, batch=False, b_size=20, b_start=0, orphan=0, **kw):
    request = request or zope.globalrequest.getRequest()
    view = api.content.get_view(
        name='folderListing',
        context=context,
        request=request,
    )
    return view(batch, b_size, b_start, orphan, **kw)
```

hvelarde:
IMO plone.api should eliminate the need for plone.app.contentlisting

jaroel:
@hvelarde I don't get what you're trying to saying? Move stuff like https://github.com/plone/plone.app.contentlisting/blob/master/plone/app/contentlisting/contentlisting.py into plone.api?

simahawk:
I agree w/ Hector: we should integrate this kind of features into plone.api.
And, I'm totally in favour of listing brains by default if we are "searching" contents.
We could have `content.list` that returns IContentListing items taken w/ `objectValues` and `content.search` that returns IContentListing items looked up via catalog.
Both can have an option `full_object` to get full objects instead of IContentListing items.

jensens:
Coming back to this one I would propose to 
- add a ``plone.api.content.list(context=context) 
- acting as a shortcut to adapt: ``plone.app.contentlisting.interfaces.IContentListing(context)``

We already have the api, but instead of having people to remember the interface location it just gives you what you need.

jensens:
I close this since it seems the solution is not accepted. reopen if i'am wrong.

Issue #157, opened 2014-01-29
user.get_permissions breaks for admin user from zope-root acl_users


thet:
when using `plone.api.user.get_permissions(user=user)` with the admin user from zope-root's acl_users folder, it breaks with a `UserNotFoundError`.
looks like the reason can be found here: https://github.com/plone/plone.api/blob/16bc0c731f5f55ac2fbed63852296c488e3c2796/src/plone/api/env.py#L39
the basic zope admin user cant be found in Plone's acl_users

jaroel:
This should be fixed in #209

PR #209, opened 2015-02-03
Support Zope users in user.adopt_user

### LINKED ISSUE #171: plone.api.user.get_current() does not deal with non-Plone uses
plone.api.user.get_current() raises an exception if the current user is defined outside the scope of the Plone site e.g. in the root of the Zope folder. get_current() should return in all cases a user object because it is common that some administrative code in add-ons is either called from a Plone site administrator and/or a Zope manager account. Right now you have write fallback code for handling this case which makes this API method worthless because the fallback code usually would implemented the get_current() functionality in a way that works in every case and not only for user accounts defined in Plone.

jaroel:
Fixes #171 and #157

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -7,6 +7,9 @@ Changelog
 - Fixed `make docs`.
   [jaroel]
 
+- Support Zope users in user.adopt_user. Fixes #171 and #157.
+  [jaroel]
+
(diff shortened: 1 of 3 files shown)


jaroel:
Please review

adamcheasley:
Looks like another network related test failure here.

zupo:
Restarting failed builds.

PR #211, opened 2015-02-03
Support a list of objects in content.delete

### LINKED ISSUE #198: content.delete should allow a list of objects
So I can replace code like
for obj in auctioncategory.objectValues('auctionsubcategory'):
    api.content.delete(obj)

with
api.content.delete(auctioncategory.objectValues('auctionsubcategory'))

jaroel:
Fixes #198

### DIFF
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -254,17 +254,24 @@ def copy(source=None, target=None, id=None, safe_id=False):
         return target[new_id]
 
 
-@required_parameters('obj')
-def delete(obj=None):
-    """Delete the object.
+@at_least_one_of('obj', 'objects')
+def delete(obj=None, objects=None):
+    """Delete the object(s).
 
-    :param obj: [required] Object that we want to delete.
+    :param obj: Object that we want to delete.
     :type obj: Content object
+    :param objects: Objects that we want to delete.
+    :type objects: List of content objects
     :raises:
         ValueError
     :Example: :ref:`content_delete_example`
     """
-    obj.aq_parent.manage_delObjects([obj.getId()])
+    if obj is not None:
+        obj.aq_parent.manage_delObjects([obj.getId()])
+    else:
+        # The objects may have different parents
+        for obj in objects:
+            delete(obj=obj)
 
 
 @required_parameters('obj')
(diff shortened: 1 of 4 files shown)


jaroel:
Please review.

src/plone/api/content.py
```diff
@@ -254,17 +254,24 @@ def copy(source=None, target=None, id=None, safe_id=False):
...
     :raises:
         ValueError
     :Example: :ref:`content_delete_example`
     """
-    obj.aq_parent.manage_delObjects([obj.getId()])
+    if obj is not None:
+        obj.aq_parent.manage_delObjects([obj.getId()])
+    else:
+        # The objects may have different parents
+        for obj in objects:
```
adamcheasley: I'm not 100% clear on why we're not just passing the list straight through to manage_delObjects here. Is there an exception when the objects have different parent objects?

jean: Yes, a parent needs to delete its child objects. So if objects in the list have different parents, each of those parents is responsible to delete the relevant child object(s). (The current code might be tweaked to gather objects per parent, so that it can call `manage_delObjects` once per parent.)

adamcheasley: Ah yes. Understood.

jaroel: @adamcheasley It even says so in the comment '# The objects may have different parent' ;)

src/plone/api/content.py
```diff
@@ -254,17 +254,24 @@ def copy(source=None, target=None, id=None, safe_id=False):
...
+            obj.aq_parent.manage_delObjects([obj.getId()])
```
mattss: Could we just pass the object back into api.content.delete, rather than having two different calls to manage_delObjects?

mattss: i.e.:

    # The objects may have different parents
    for obj in objects:
        api.content.delete(obj=obj)

adamcheasley: That seems slightly more elegant to me.

jaroel: I've changed this.

zupo:
Looks good, but needs a rebase off of latest master so it can be merged.

gforcada:
rebased

Issue #215, opened 2015-02-08
Optional request arguments


jaroel:
As a plone.api consumer I want a small as possible function signatures.
So
```python
view = api.content.get_view(name='plone', context=context)
```
instead of
```python
view = api.content.get_view(name='plone', context=context, request=context.REQUEST)
```

jaroel:
Won't fix. It's way too much hassle to deal with the edge cases. See @davisagli 's comment in #216.

PR #216, opened 2015-02-08
Make request arguments optional.

jaroel:
Implements #215

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 1.3.3 (unreleased)
 ------------------
 
+- Made the request argument optional.
+  [jaroel]
+
 - Make Products.Archetypes dependency optional. Fixes #197
   [jaroel]
 
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -358,11 +358,7 @@ To get a :class:`BrowserView` for your content, use :meth:`api.content.get_view`
 
     from plone import api
     portal = api.portal.get()
-    view = api.content.get_view(
-        name='plone',
-        context=portal['about'],
-        request=request,
-    )
+    view = api.content.get_view(name='plone', context=portal['about'])
 
 .. invisible-code-block: python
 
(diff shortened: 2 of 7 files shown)


jaroel:
Please review.

zupo:
Ugh, I think we had it like this already but then people complained that we should not use thread locals / global request. Would like to hear more opinions on this one before we merge it ...

adamcheasley:
IMO as long as we still have the option to pass in the request, then this should be fine.

jaroel:
That's still there. I just don't want to have to pass request, if zope.globalrequest.getRequest would do :)

jaroel:
@adamcheasley Please merge if it's acceptable.

zupo:
@davisagli can you please take a look at this one? I'm sure there were people against using the global request and you name pops up in my head as one such potential person.

zupo:
For what is worth, the Pyramid folks are *strongly* against using global requests: http://docs.pylonsproject.org/projects/pyramid/en/1.0-branch/narr/threadlocals.html

"You probably should almost never use the get_current_request() or get_current_registry() functions, except perhaps in tests. "

davisagli:
There's no doubt that avoiding thread locals is cleaner. The main arguments against them are:
1. They assume your code is always going to be run by a threaded server (probably true for zope, but not necessarily for people experimenting with async approaches) with only one request being served per thread.
2. They're not re-entrant (so, for example, one module being published by the Zope publisher can't call to publish a subrequest without fiddling with the thread locals to make sure they're restored afterward).
3. They make your code harder to test and run outside of the web framework, because you have to remember to set up and tear down the thread locals.

On the flip side, it's a nicer experience for the developer if they don't have to pass the request around everywhere.

What if we made a helper like this:

    from Testing.makerequest import makerequest
    from zope.globalrequest import getRequest
    import Zope2
    def get_request():
        request = getRequest()
        if request is None:
            request = makerequest(Zope2.app()).REQUEST
        return request
and use that in plone.api when it wants an implicit request? That way at least it doesn't break in tests and zopectl run scripts, addressing objection 2 above. (Might need to add some marker interfaces to the request from makerequest.)

But I think for the cases shown here, get_view and show_message, it's probably better to be explicit about it, since they are both actions that don't make much sense without a request. Actually in the case of show_message I wish it was just a method on the request so I didn't have to import anything, even from plone.api.

jaroel:
Fair enough :)

PR #224, opened 2015-04-25
transition() does _not_ return the new state. 

zopyx:
The implementation always returns None. So the documentation is wrong here.


gforcada:
Good catch!

PR #229, opened 2015-05-12
Use more generic 'noqa' instead of 'noPEP8'.

tisto:


gforcada:
You want to merge into ``deprecate-rtd`` branch or into master?

PR #238, opened 2015-06-09
Don't swallow exceptions in get_view

### LINKED ISSUE #230: Replace bare try..except
The bare try..except's must be replaced with catching only the specific exception, so the underlying errors are shown.

For example: if one has an error in a view's \__init\__, get_view will say the view doesn't exist, but in reality the view raises AttributeError.

jaroel:
Fixes #230

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 1.3.3 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


jaroel:
The build now passes. Something with timeouts. https://travis-ci.org/plone/plone.api/builds/66118518

adamcheasley:
Tests ran fine locally. Code looks good. Merging.

PR #239, opened 2015-06-09
Fix external links to docs.

### LINKED ISSUE #202: Move references  from api.plone.org to docs.plone.org?
I quickly grepped and saw that there are references to api.plone.org, should they be moved to docs.plone.org?

I see that #130 was created for moving from RTD to api.plone.org, but given that docs.plone.org is in a healthy and good shape, should we move everything there?

Just wondering...

jaroel:
Fixes #202

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ Source Code and Contribution
     :target: https://coveralls.io/r/plone/plone.api?branch=master
 
 Contributors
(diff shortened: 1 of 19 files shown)


gforcada:
@hvelarde there is a link that should have not been changed... please review the changes before merging, that's what I was actually doing and was about to push...

Issue #228, opened 2015-05-07
send_email does not work with PrintingMailHost w/o MailHost configured


jensens:
In development scenarios Products.PrintingMailHost helps a lot. 
But ``send_email`` does not work with it unless a Mailhost is configured in the registry. In dev, when one trashes all the time the database this is annoying.

The configuration check is great, but we may want to disable it if Plone is in debug-mode?

zupo:
+1

gforcada:
:+1: maybe checking specifically for P.PrintingMailHost being around would be cleaner? I can imagine someone trying to send real emails from a debug instance (maybe crazy, but worse things can be seen on the wild...)

jensens:
+1 @gforcada - good point

jaroel:
What needs to be done, and who is going to do this?

jensens:
@jaroel I'd say @gforcada for checking specifically for ``Products.PrintingMailHost`` and when there disabling the check is the task. An easy task. I can do this but have no time in next future to do so. Any volunteers welcome!

jaroel:
I have no idea on how I can test this without un-patching and patching the PrintingMailHost patches.

jaroel:
This is resolved in #244

PR #244, opened 2015-06-14
Support Products.PrintingMailHost.

jaroel:
Implements #228.
No idea on how to test this.

### DIFF
--- a/src/plone/api/tests/test_portal.py
+++ b/src/plone/api/tests/test_portal.py
@@ -303,7 +303,8 @@ class TestPloneApiPortal(unittest.TestCase):
         self.assertEqual(msg['From'], 'Registry <reg@example.org>')
 
     def test_send_email_with_printingmailhost(self):
-        """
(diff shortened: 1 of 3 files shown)


gforcada:
For testing, injecting a global ``PRINTINGMAILHOST_ENABLED`` before calling ``get`` could be enough?

jaroel:
@gforcada yes, it does :) Thanks!

src/plone/api/tests/test_portal.py
```diff
@@ -302,6 +302,35 @@ def test_send_email_with_config_in_registry(self):
         msg = message_from_string(self.mailhost.messages[0])
         self.assertEqual(msg['From'], 'Registry <reg@example.org>')
 
+    def test_send_email_with_printingmailhost(self):
+        """
+        """
```
gforcada: An empty docstring is quite useless, we could just remove it.

jaroel:
This should just fine, could someone merge this?

PR #251, opened 2015-06-23
Add an optional 'default' argument to plone.api.content.get_state.

### LINKED ISSUE #246: plone.api.content.get_state should handle objects without a workflow definition.
Products.CMFCore.WorkflowCore.WorkflowException is raised when there is no workflow definition for the object.
This make the current abstraction leaky, ie the client code needs to handle the exception and we therefor cannot replace the underlying workflow engine.

The implementation should check that there really is no workflow is associated with the content object, instead of just catching the exception. WorkflowException is too generic for this. Catching it will mask other errors and prevents logging them.

We should handle this situation and return None instead to make the client code cleaner.
Here is an example for the client code:

The new approach gives us code that looks like:
```python
state = api.content.get_state(context=my_image)
if state is None:
    # Handle this somehow
else:
    print "state: ", state
```
or
```
state = api.content.get_state(context=my_image)
print "state: ", state if state else 'Unknown'
```

Where in the old situation client code looks like:
Handle the exception
```python
from Products.CMFCore.WorkflowCore import WorkflowException

try:
    state = api.content.get_state(context=my_image)
except WorkflowException:  # this swallows all other WorkflowExceptions too.
   # Handle this somehow
else:
    print "state: ", state
```

jaroel:
The default value is returned when there is no workflow defined for the object, so the client code doesn't need to catch WorkflowException.

Closes #246

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 1.3.3 (unreleased)
 ------------------
 
+- plone.api.content.get_state now allows for an optional default value.
+  This is used when no workflow is defined for the object. Refs #246
+  [jaroel]
+
 - plone.api.portal.get_registry_record now suggests look-alike records when no records is found. Refs #249.
   [jaroel]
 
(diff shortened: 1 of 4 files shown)


jaroel:
I'll rebase this if when tests pass.

jaroel:
This can be reviewed and merged.

PR #259, opened 2015-07-14
use groupid instead of groupname

pbauer:
This is change in the api and should only be merged after a 1.3.3 is released.


pbauer:
@adamcheasley I said "only be merged after 1.3.3 is released". Not before.

adamcheasley:
@pbauer Sorry, I missed your comment.

pbauer:
np

PR #263, opened 2015-07-14
Add test to demonstrate issue

hvelarde:


adamcheasley:
It appears the tests have failed due to the fact that in plone 5, you don't actually get an unauthorized raised. In that case, the function raises an `InvalidParameterError`

jaroel:
I've fixed the actual problem in #264.

hvelarde:
no, the point is a user with Member role is not allowed to create content; as you can see the test passes in other Plone versions but 5.0.

hvelarde:
the issue happens with Dexterity-based content types; test could fail under Plone 4 if we were using plone.app.contenttypes.

jaroel:
Yep, this is because plone.dexterity.content.invokefactory does FTI checks before permission checks.

jaroel:
Please file a bug report with plone.dexterity ;) :+1:

hvelarde:
@jaroel why don't we merge the test?

jaroel:
We'll need to add a conditional dependency to plone.dexterity >= 2.3.3

jaroel:
I don't see how I can do this.
We either have plone.dexterity, or we don't. This needs to be in the KGS.

Issue #160, opened 2014-04-03
Wrong error message when creating a content type object without permissions


hvelarde:
If you have a folder created, let's say, by the site admin, and then you try to add an object inside the folder with an user that doesn't have permissions, you get a **Disallowed subobject type** error instead of a permission one.

this is very confusing.

```python
Error in test test_view_synchronise (my.project.tests.test_synchronise.ContentTypeTestCase)
Traceback (most recent call last):
  File "/home/hvelarde/buildout.python/parts/opt/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/home/hvelarde/simplesconsultoria/my.project/src/my/project/tests/test_synchronise.py", line 50, in test_view_sync
    counter = view.create_content()
  File "/home/hvelarde/simplesconsultoria/my.project/src/my/project/sync/synchronise.py", line 73, in create_content
    o = self.create_mytype(base_folder['container'], par)
  File "/home/hvelarde/simplesconsultoria/my.project/src/my/project/sync/synchronise.py", line 113, in create_mytype
    folder, type='MyType', id=oId, title=title)
  File "<string>", line 2, in create
  File "/home/hvelarde/.buildout/eggs/plone.api-1.1.0-py2.7.egg/plone/api/validation.py", line 70, in wrapped
    return f(*args, **kwargs)
  File "<string>", line 2, in create
  File "/home/hvelarde/.buildout/eggs/plone.api-1.1.0-py2.7.egg/plone/api/validation.py", line 131, in wrapped
    return f(*args, **kwargs)
  File "/home/hvelarde/.buildout/eggs/plone.api-1.1.0-py2.7.egg/plone/api/content.py", line 84, in create
    "{2}".format(type, '\n'.join(sorted(types)), e.message)
InvalidParameterError: Cannot add a 'MyType' object to the container.
Allowed types are:
MyType
Disallowed subobject type: MyType
```

jaroel:
What needs to be done, and who is going to do it?

adamcheasley:
I think we have a pretty clear plan. If I manage to get #204 done in the next couple of weeks, I will take a look at this one too.

adamcheasley:
What exception would be best to raise here? AccessControl.Unauthorized?

jaroel:
Works just fine on master:

```python
>>> with api.env.adopt_roles(roles='Manager'): my_folder = api.content.create(type='Folder', id='my_folder', container=api.portal.get())
>>> my_folder
[...]

PR #264, opened 2015-07-15
Fix plone.api.content.create in Plone 5.

jaroel:
getLocallyAllowedTypes is no longer available by default. allowedContentTypes is and it does the same thing.

Refs #160


adamcheasley:
Yes! 10 lines down to 1. Excellent!

PR #267, opened 2015-07-30
Make object_provides work also on tuple

### LINKED ISSUE #266: object_provides should accept tuple
If the passed value for object provides is a tuple you get this error:
```
(Pdb) type(self._analyzed_type_ifaces)
<type 'tuple'>
(Pdb) api.content.find(object_provides=self._analyzed_type_ifaces)
*** AttributeError: 'tuple' object has no attribute '__identifier__'
```

This is solved by casting the tuple to list:
```
(Pdb) api.content.find(object_provides=list(self._analyzed_type_ifaces))
[<Products.ZCatalog.Catalog.mybrains object at ...>, 
   ...
]
```

I would do this somewhere around this line:
 - https://github.com/plone/plone.api/blob/46a6736a56c9b0f8075fafd5882ce72481c4da70/src/plone/api/content.py#L552

ale-rt:
Fixes #266

### DIFF
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -549,12 +549,11 @@ def find(context=None, depth=None, **kwargs):
     # Convert interfaces to their identifiers
     object_provides = query.get('object_provides', [])
     if object_provides:
-        if not isinstance(object_provides, list):
+        if not isinstance(object_provides, (list, tuple)):
             object_provides = [object_provides]
-        for k, v in enumerate(object_provides):
-            if not isinstance(v, basestring):
-                object_provides[k] = v.__identifier__
-        query['object_provides'] = object_provides
+        query['object_provides'] = [
+            getattr(x, '__identifier__', x) for x in object_provides
+        ]
 
     # Make sure we don't dump the whole catalog.
     catalog = portal.get_tool('portal_catalog')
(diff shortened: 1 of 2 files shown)


src/plone/api/content.py
```diff
@@ -549,12 +549,11 @@ def find(context=None, depth=None, **kwargs):
     # Convert interfaces to their identifiers
     object_provides = query.get('object_provides', [])
     if object_provides:
-        if not isinstance(object_provides, list):
+        if not isinstance(object_provides, (list, tuple)):
```
adamcheasley: Why not make the function work with all iterable types? This would work:

```
import collections
if isinstance(object_provides, collections.Iterable):
```

ale-rt: Yeah, you are right, this is not the best possible solution.

Anyway checking for collections.Iterable leads you to problem in case of dicts and strings:
```
>>> a = {}
>>> b =''
>>> import collections
>>> isinstance(a, collections.Iterable)
True
>>> isinstance(b, collections.Iterable)
True
```

This would lead to additional "if"s :)

adamcheasley: Ah true. Fair point.

jensens:
please add a note to the changelog

ale-rt:
@jensens done ;)

PR #273, opened 2015-08-26
Add option to skip checking linkintegrity to plone.api.content.delete

pbauer:
Please review (works only with https://github.com/plone/plone.app.linkintegrity/pull/23). 

The main questions:
- Is it ok that the default for ``check_linkintegrity`` is ``False`` 
- Is raising the exception ``plone.app.linkintegrity.exceptions.LinkIntegrityNotificationException`` the right way to deal with breaches?
- It only works in Plone 5 (current coredev).

### DIFF
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -34,7 +34,7 @@ else:
 
 # Old linkintegrity (Plone <= 5.0b4) or new (Plone > 5.0b4)
 linkintegrity_version = get_distribution('plone.app.linkintegrity').version
-if parse_version(linkintegrity_version) >= parse_version('3.0'):
+if parse_version(linkintegrity_version) >= parse_version('3.0.dev0'):
     NEW_LINKINTEGRITY = True
 else:
     NEW_LINKINTEGRITY = False
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -3,7 +3,9 @@
 
 from Products.CMFCore.WorkflowCore import WorkflowException
 from copy import copy as _copy
+from pkg_resources import DistributionNotFound
 from pkg_resources import get_distribution
+from pkg_resources import parse_version
 from plone.api import portal
 from plone.api.exc import InvalidParameterError
 from plone.api.validation import at_least_one_of
@@ -19,20 +21,20 @@ from zope.container.interfaces import INameChooser
 from zope.interface import Interface
 from zope.interface import providedBy
 
-import pkg_resources
 import random
 import transaction
 
 try:
-    pkg_resources.get_distribution('Products.Archetypes')
-except pkg_resources.DistributionNotFound:
+    get_distribution('Products.Archetypes')
+except DistributionNotFound:
     class IBaseObject(Interface):
         """Fake Products.Archetypes.interfaces.base.IBaseObject"""
 else:
     from Products.Archetypes.interfaces.base import IBaseObject
 
 # Old linkintegrity (Plone <= 5.0b4) or new (Plone > 5.0b4)
-if get_distribution('plone.app.linkintegrity').version >= 3.0:
+linkintegrity_version = get_distribution('plone.app.linkintegrity').version
+if parse_version(linkintegrity_version) >= parse_version('3.0'):
     NEW_LINKINTEGRITY = True
 else:
     NEW_LINKINTEGRITY = False
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -8,6 +8,8 @@ from plone.api.exc import InvalidParameterError
 from plone.api.validation import at_least_one_of
 from plone.api.validation import mutually_exclusive_parameters
 from plone.api.validation import required_parameters
+from plone.app.linkintegrity.exceptions import \
+    LinkIntegrityNotificationException
 from plone.app.uuid.utils import uuidToObject
 from plone.uuid.interfaces import IUUID
 from zope.component import getMultiAdapter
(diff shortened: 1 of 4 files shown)


adamcheasley:
- I would probably vote for the default to to be `True`. 
- I think raising the exception is the correct behaviour. 
- With regards to only working in Plone 5, what would happen to Plone 4 compatibility? Would I get an ImportError at runtime or would I just not be able to use this part of the function?

hvelarde:
we can solve compatibility issue by not allowing `check_linkintegrity=False` in Plone 4.

pbauer:
Note: The new linkintegrity-stuff is not part of in 5.0b4. Until Plone 5.0b5 ``plone.api.content.delete`` will always throw ``LinkIntegrityNotificationException`` (unless you use dexterity without plone.app.referenceablebehavior).
 
My idea: 
We add ``check_linkintegrity=True`` for all versions: For < 5.0b5 it would wrap ``obj.aq_parent.manage_delObjects`` in a try/except clause that swallows the ``LinkIntegrityNotificationException`` if check_linkintegrity is True. For >= 5.0b5 it would use to logic of the current pull-request. The main problem ist that the code would not get more beautiful but maybe that is what apis are for, wrapping ugly code in nice code.

adamcheasley:
@pbauer Sounds like a good idea to me!

pbauer:
I reworked it as proposed. All tests are green. Please review.

src/plone/api/tests/test_content.py
```diff
@@ -610,6 +616,81 @@ def test_delete_multiple(self):
         assert 'copy_of_about' not in container
         assert 'about' not in container['events']
 
+    def test_delete_ignore_linkintegrity(self):
+        """Test deleting a content item with a link pointed at it."""
+        self._set_text(self.team, '<a href="contact">contact</a>')
+        # Delete the contact page
+        api.content.delete(self.contact, check_linkintegrity=False)
+        assert 'contact' not in self.portal['about'].keys()
```
hvelarde: shouldn't this be `self.assertNotIn( 'contact', self.portal['about'].keys())`?

src/plone/api/tests/test_content.py
```diff
@@ -610,6 +616,81 @@ def test_delete_multiple(self):
...
+    def test_delete_check_linkintegrity(self):
+        """Test deleting a content item with a link pointed at it."""
+        if not IBaseContent.providedBy(self.team) or not NEW_LINKINTEGRITY:
+            # This test only makes sense with Archetypes or new Linkintegrity
+            return
+        self._set_text(self.team, '<a href="contact">contact</a>')
+        # Delete the contact page
+        with self.assertRaises(LinkIntegrityNotificationException):
+            api.content.delete(self.contact)
+        assert 'contact' in self.portal['about'].keys()
```
hvelarde: same here

src/plone/api/tests/test_content.py
```diff
@@ -610,6 +616,81 @@ def test_delete_multiple(self):
...
+        # Delete linked pages
+        api.content.delete(
+            objects=[self.blog, self.contact],
+            check_linkintegrity=False)
+        self.assertNotIn('contact', self.portal['about'].keys())
+        self.assertNotIn('blog', self.portal.keys())
+
+    def test_delete_with_internal_breaches(self):
+        """Test deleting multiple with internal linkintegrity breaches."""
+        if not IBaseContent.providedBy(self.team) or not NEW_LINKINTEGRITY:
```
hvelarde: same here

src/plone/api/tests/test_content.py
```diff
@@ -610,6 +616,81 @@ def test_delete_multiple(self):
...
+            api.content.delete(objects=[self.blog, self.contact])
+        if NEW_LINKINTEGRITY:
+            # In the old implementation of linkintegrity the items are
+            # still gone during this request.
+            self.assertIn('contact', self.portal['about'].keys())
+            self.assertIn('blog', self.portal.keys())
+
+    def test_delete_multiple_ignore_linkintegrity(self):
+        """Test deleting multiple items ignoring linkintegrity-breaches."""
+        if not IBaseContent.providedBy(self.team) or not NEW_LINKINTEGRITY:
```
hvelarde: is better to use `unittest.skipIf` here

pbauer: In the decorator ``@skipIf`` I have no access to ``self``, so I cannot test of ``self.team`` is Archetypes or Dexterity. 
Instead I could after the imports if ``plone.app.contenttypes`` is installed which would mean it is Dexterity.

hvelarde:
@pbauer just one more thing: can we squash all commits into one?

src/plone/api/content.py
```diff
@@ -16,18 +21,24 @@
...
-import pkg_resources
 import random
 import transaction
 
 try:
-    pkg_resources.get_distribution('Products.Archetypes')
-except pkg_resources.DistributionNotFound:
+    get_distribution('Products.Archetypes')
+except DistributionNotFound:
     class IBaseObject(Interface):
```
hvelarde: if you do this all objects will provide this interface, right? why not add a boolean variable stating whether you have or not Archetypes installed?

pbauer: I did not add this but it is a good question. When Products.Archetypes is not installed  ``IBaseObject.providedBy(obj)`` would then always be true, would it not?
@jaroel Can you have a look at that (since you added it...)?

jaroel: Missed this somehow, sorry.

This won't happen, because this IBaseObject is a different class than the one in Products.Archetypes.
It is probably clearer when you look at the .\__identifier__ (or the dotted name): `Products.Archetypes.interfaces.base.IBaseObject` or `plone.api.content.IBaseObject`.

'obj' may provide 'Products.Archetypes.interfaces.base.IBaseObject', but will never 'plone.api.content.IBaseObject'.
Unless, of course, people start doing 'zope.interface.alsoProvides(obj, plone.api.content.IBaseObject)'. Please do not do this :)

PR #274, opened 2015-09-03
Add section on squashing commits into the conventions

hvelarde:


hvelarde:
this probably needs a better redaction or examples; please feel free to comment and improve.

adamcheasley:
Just made a small grammar correction, but otherwise looks good. I don't think we need too many examples, as a contributor should have at least some git knowledge. Maybe we could add a link to the git documentation?

hvelarde:
talking on squashing... I already squashed you corrections

gforcada:
Maybe a link to https://github.com/plone/buildout.coredev/blob/5.0/docs/git.rst (wherever that's published)?

PR #275, opened 2015-09-07
Fix issues after use_email_as_login was moved to registry

pbauer:


esteele:
Looks good. Thank you.

PR #276, opened 2015-09-11
Fix content.find with both path and depth

mauritsvanrees:
If you specify a depth and a path, until now we ignored the path. Not nice.

I discovered this while developing on a client project. I did `content.find(path='/plone/folder', depth=1)` and it returned all items in the Plone site root instead of in `/plone/folder`.

The alternative is to do `content.find({'path': {'query': '/plone/folder', 'depth': 1}})`, but I have always find this ugly.

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog
 1.4.6 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
Jenkins coredev 5.0 running here: http://jenkins.plone.org/job/pull-request-5.0/296/

mauritsvanrees:
Jenkins tests have passed. Please review.

PR #277, opened 2015-09-13
No need for unittest2

gforcada:
Since python 2.7 unittest2 was already merged with unittest.

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -8,6 +8,9 @@ Changelog
   is no longer ignored then.
   [maurits]
 
+- Remove unittest2 dependency.
+  [gforcada]
+
 
 1.4.5 (2015-09-09)
 ------------------
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,6 @@ setup(
             'plone.app.testing',
             'plone.indexer',
             'plone.registry',
-            'unittest2',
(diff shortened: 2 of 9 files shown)


jensens:
@gforcada please add a line to the changelog

tests: http://jenkins.plone.org/job/pull-request-5.0/308/

mauritsvanrees:
Isn't plone.api master still meant for older versions of Plone too?  Plone 4.2 and 4.3 still support Python 2.6, so plone.api should too. Of course unittest2 can then be added conditionally to the install_requires, but the imports would need to stay the same if I see it correctly.

mauritsvanrees:
We test master on Travis with Plone 4.2 and Python 2.6.

Oh, but on this pull request Travis passes... Was unittest2 merged into Python 2.6 as well maybe?

gforcada:
@vangheem good points, and no idea about why it passes... if at all I would prefer to leave the code cleaner than more involved just for a stupid dependency, so let's not merge it until we are sure we can do so.

mauritsvanrees:
i guess we are simply not using any unittest2 features. That would explain it.

gforcada:
Then it is safe to merge... let me resolve the conflicts.

mauritsvanrees:
Agreed.

PR #278, opened 2015-09-16
Add Plone 5 only tests to make sure the new registry lookups work.

tisto:


tisto:
I'm not sure if it makes sense to have those kind of tests here. Though, I needed to test things before they go into the Plone 5 upgrade guide. Since we can't use plone.api in the core, I can't add them to CMFPlone.

gforcada:
What about adding them in plone.app.upgrade? If feels *wrong* to have them here as they are not related at all with plone.api (well, they use it of course).

And if you say that they are meant for the upgrade guide ... p.a.upgrade would be my suggestion. There's already tests on it.

src/plone/api/tests/test_portal_get_registry.py
```diff
@@ -0,0 +1,328 @@
...
+    def test_editing_settings(self):
+        keys = [
+            'plone.available_editors',
+            'plone.default_editor',
+            'plone.ext_editor',
+            'plone.enable_link_integrity_checks',
+            'plone.lock_on_ttw_edit',
+        ]
+        for key in keys:
+            raised = False
```
hvelarde: IMO, this should be better written:

```python
from plone.api.exc import InvalidParameterError
...
try:
    api.portal.get_registry_record(key)
except InvalidParameterError:
    self.fail('"{}" not found in registry.'.format(key))
```

hvelarde:
I also think this makes no sense in here.

jaroel:
Please put this in plone.app.upgrade :)

PR #279, opened 2015-09-22
Add autopep8 to the convention docs

mauritsvanrees:
Plus run it on our own code for good measure.

### DIFF
--- a/docs/contribute/conventions.rst
+++ b/docs/contribute/conventions.rst
@@ -88,6 +88,61 @@ This package follows all rules above, `check out the source
 <https://github.com/plone/plone.api/tree/master/src/plone/api>`_ to see them
 in action.
 
+autopep8
+--------
+
+Making old code pep8 compliant can be a lot of work.  There is a tool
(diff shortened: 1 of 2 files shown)


jensens:
After this stucks a while I ping the doc-heros @polyester and @svx for a comment.

My opinion: I'am not 100% sure if the conventions document is the right place for tool explanations.
I'am totally +1 to have this documented in the plone docs.

mauritsvanrees:
I have put it in the conventions document, because this is a tool that assist in keeping to one of the conventions.

But anywhere in the Plone docs is fine with me. Or only a link to https://pypi.python.org/pypi/autopep8

polyester:
as long as it's clear you run this *only* on .py files, and **never** on .rst files. Can't hurt repeating that.

There's no real 'optimal' place in the docs yet, but we point to the plone.api conventions elsewhere as well as Good Practice & Things You Should Do, so it's fine here.

jensens:
travis fails for 5.0 after merge, but seems unrelated to this merge, see #281

PR #284, opened 2015-11-19
fixes #283 portal.send_email does not respect transaction aborts

### LINKED ISSUE #283: portal.send_email does not respect transaction aborts
plone.api.portal.send_email uses immediate=True. This means that the email is sent multiple times if a ConflictError occurs.
Unless explicitly specified, email should be send at transaction boundary and/or using a mail queue.

https://github.com/plone/plone.api/blob/master/src/plone/api/portal.py#L177
http://docs.plone.org/develop/plone/misc/email.html

jensens:
enable sane default behavior respecting transactions

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -10,7 +10,8 @@ New:
 
 Fixes:
 
-- *add item here*
+- #283 portal.send_email does not respect transaction aborts
+  [jensens]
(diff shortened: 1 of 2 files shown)


jensens:
http://jenkins.plone.org/job/pull-request-5.0/558/

Issue #269, opened 2015-08-16
Allow [sg]et_registry_record to get an optional interface


gforcada:
I add my records to the registry via a Interface, then to retrieve them back I use:

```
registry = getUtility(IRegistry)
records = registry.forInterface(IMyInterface)
records.my_register_record = 'something'
```

It would be great if plone.api allows me to do the same:

```
get_registry_record('my_register_record', interface=IMyInterface)
```

Sounds good/desirable? Wording of course up to debate :-)

hvelarde:
+1 awesome!

pcdummy:
+1, its very nice to expose registry settings on a view.

PR #286, opened 2015-11-26
Allow to set/get registry settings from an interface

gforcada:
Fixes:
https://github.com/plone/plone.api/issues/269

### DIFF
--- a/src/plone/api/portal.py
+++ b/src/plone/api/portal.py
@@ -18,6 +18,7 @@ from zope.component import getUtility
 from zope.component import providedBy
 from zope.component.hooks import getSite
 from zope.globalrequest import getRequest
+from zope.interface.interfaces import IInterface
 
 import pkg_resources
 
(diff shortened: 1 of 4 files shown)


src/plone/api/portal.py
```diff
@@ -281,19 +309,64 @@ def get_registry_record(name=None):
...
+        raise InvalidParameterError(
+            u"The interface parameter has to derive from "
+            u"zope.interface.Interface"
+        )
+
     registry = getUtility(IRegistry)
-    if isinstance(name, str):
+
+    if interface is not None:
+        records = registry.forInterface(interface)
```
adamcheasley: This appears to be an exact copy of the code above. Would it not be better to call `get_registry_record` here?

gforcada: I was thinking on refactoring it as an external method, but calling ``get_registry_record`` sounds good as well

gforcada: Done!

mauritsvanrees:
@adamcheasley When you merge a pull request, can you first check that the package is added in the checkouts.cfg of the relevant plone-coredev branch? Thanks.
I have now done that: https://github.com/plone/buildout.coredev/commit/cf07d530ab5a8f66729a02cd9836080831885a4b
Sorry if I was too fast and you were just about to do that. :-)

adamcheasley:
No problem. I was there ready to commit, but by the time I'd hit the button, you'd already done it!
Thank you.

mauritsvanrees:
Perfect.

gforcada:
Thanks for merging!

PR #296, opened 2016-01-13
WIP Update docstring on send_email

gforcada:
Although a string is expected (and only if it's unicode is encoded),
passing a python's email object works as well.

This avoids having to change send_email signature to support every field that could be used on an email,
like attachments, reply_to, CC, etc etc

### DIFF
--- a/docs/portal.rst
+++ b/docs/portal.rst
@@ -132,8 +132,6 @@ Send E-Mail
 
 To send an e-mail use :meth:`api.portal.send_email`:
 
-.. Todo: Add example for creating a mime-mail
-
 .. invisible-code-block: python
 
     # Mock the mail host so we can test sending the email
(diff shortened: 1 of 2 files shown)


gforcada:
One question remains though, should we allow that a user could just pass an email object as well? So:

```
my_email_obj = email.EMail()  # pseudocode
api.portal.send(my_email_obj)
```

Or should we still force the user to pass all the required parameters?

Opinions?

jensens:
:+1:

gforcada:
I added a MIME example, as for how to these emails, I'm thinking that we either remove the decorator and do the checks within the function or create a new method ``api.portal.send_email_object``.

I can't decide which one is less worse (I don't really like much any of them).

thet:
I do not understand the problem... your example shows how you want to support passing an email instance, ``message`` being the email object:

```
api.portal.send_email(
    recipient="bob@plone.org",
    sender="noreply@plone.org",
    subject="Trappist",
    body=message,
)
```

The signature of send email is like follows and does only require ``recipient``, ``subject`` and ``body``, not enforcing any of these being a string:

```
@required_parameters('recipient', 'subject', 'body')
def send_email(sender=None, recipient=None, subject=None, body=None,
               immediate=False):
```

So, what you want is already supported. Maybe I'd add an assert statement to enforce basestring (or unicode) or an instance of this email object you're referring to - if having asserts is plone.api practice...

Or did I miss something?

thet:
+1, by the way.

gforcada:
@thet the example I posted (above, see below) is not possible right now:

```
my_email_obj = email.EMail()  # pseudocode
api.portal.send(my_email_obj)
```

i.e. if you are already building an email object, it can be seen as a duplication to have to pass sender, recipient and subject, while they are already on the email object.

So what I was proposing was to be able to do a plain email, where you pass all the fields to api.portal.send, or you just pass a single email object. Given the current decorators api.portal.send has it is not possible at all.

thet:
ou kay. so the problem is, that the validator decorators in https://github.com/plone/plone.api/blob/master/src/plone/api/validation.py does not support "either one of this set of parameters or one of the other".

this could be solved, if the ``required_parameters`` would be extended to catch ``MissingParameterError`` and revalidate against another set of parameters, so that you can write:

```
@required_parameters('body')
@required_parameters('recipient', 'subject', 'body')
def send_email(...):
    ...
```

Oh, it would be nice, if the required parameters would support typing too, like:
``@required_parameters('body', types={'body': email.Email})``

gforcada:
@thet any of these solutions seem elegant enough, so +1 for me.

gforcada:
So we can merge this PR and work on another one to add either solution seems best to support this?

PR #304, opened 2016-01-31
Cleanup

gforcada:


gforcada:
And monitored: http://jenkins.plone.org/view/Pkgs/job/package-plone.api/

gforcada:
So it currently fails due to using AccessControl < 3.0 on Plone 4.2.

Do we want to maintain Plone 4.2 compatibility up until ... ?

jensens:
if we skip Plone 4.2 compatibility we need to increase the major version number.

gforcada:
Skipping 2.6 compatibility then is a no-go for Plone 4.3 (as we still currently support it).

I will revert those specific changes. It's a bit lame to remove Plone 4.2 support only due to some QA changes.

jensens:
ah, sure. I wonder if anybody is using Plone 4.3 with Python 2.6 anyway ;) but its indeed a thing to support.

gforcada:
Ok, remove those decorators, it should get green :-)

hvelarde:
Python 2.6 was first released in 2008 and, since 2013, is no longer maintained; I don't see any reason to keep maintaining compatibility with it; Plone 4.2 runs under Python 2.7 also.

gforcada:
@hvelarde if Plone says that we support python 2.6 on Plone 4.3, then we either make an statement that this is no longer true or we maintain its support (that's why jenkins is still running 2.6 tests as well).

PR #306, opened 2016-02-10
Corner case: only exit transitions

gforcada:
If a workflow has a state where no transition goes back to it,
an initial state has to be,
then without this check the line below this patch fails as the
state id is not found on the dictionary.

Should I add a test for it? Looking at the current tests it seems that it would be too much for such a small change.

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -25,6 +25,10 @@ Fixes:
 - Cleanup code to match Plone's style guide.
   [gforcada]
 
+- Fix corner case on content.transition code: if a transition only has
+  exit transitions and no transition goes back to it ``find_path`` will fail.
+  [gforcada]
+
 1.4.11 (2016-01-08)
 -------------------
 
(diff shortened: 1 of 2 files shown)


adamcheasley:
I think it's fine without a test. I will merge once Jenkins finishes.

gforcada:
@adamcheasley thanks for the promptly response!

The 4.3 job fails because plone.api is not on buildout.coredev sources.cfg, so I guess it's safe to ignore it. Let's wait for the 5.0 job though.

hvelarde:
I would like to see that documented in the code: in a few weeks nobody will remember why we have that condition there.

gforcada:
@hvelarde sure I can add that... the method(s) needs not just a comment on that line only though

gforcada:
@hvelarde docs added

hvelarde:
awesome!

gforcada:
@adamcheasley would you mind merging if it's ok? :-)

PR #308, opened 2016-02-10
Handle automatic transitions

gforcada:
If content.transition is called with a to_state parameter,
a list of transitions that make possible to move from the object original
state to the final desired state is compiled with ``_wf_transitions_for``.

If that list of transitions contain any automatic transition and is triggered
the iteration over that list of transitions will be broken as the
automatic transition will have already happened.

This commit handles this case.

Again, just like #306 making a test for this seems not trivial at all :-/

Is there a way to create workflows on the fly easily?

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -29,6 +29,9 @@ Fixes:
   exit transitions and no transition goes back to it ``find_path`` will fail.
   [gforcada]
 
+- Handle automatic transitions on api.content.transition.
+  [gforcada]
+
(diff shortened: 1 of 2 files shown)


gforcada:
Anyone up for reviewing/merging? :-)

hvelarde:
we need a rebase here

gforcada:
@adamcheasley or anyone else, any chances to get this merged?

gforcada:
@hvelarde thanks!

PR #310, opened 2016-02-18
Remove commit on integration test

gforcada:
On integration tests one is not supposed to make commits, that would break test isolation.

Either remove the commit or turn them into functional tests


gforcada:
@do3cc seems that plone.api is one to be blamed with https://github.com/plone/plone.testing/pull/19

PR #319, opened 2016-08-31
fix #99 for Archetypes content

### LINKED ISSUE #99: api.content.create replaces archetype-content title with site title during Plone site creation
How to reproduce:
1. Use `plone.api.content.create` to create content in "setuphandlers.py" (or custom GS profile configuration step).
2. Create a new Plone site and select the profile in question to be installed for the new site.

Then

Each new Archetype-based content object created with plone.api.content.create will have site title as its title.

Why
1. plone.api.content.create calls Archetypes' processForm-method
2. processForm-method tries to re-fill content fields from request
3. title-widget finds a title field from the request (because the default form for creating a new Plone site has a title field)

Our local fix is to pop "title" key from the request.form before running code with plone.api.content.create.

But note that this is not limited to the site creation from but any values in request.form may cause similar surprises with api.content.create.

If it's not a feature to read field values for plone.api.content.create from request, then maybe request.form should somehow be cleared before creating Archetypes-content and restored afterwise.

gotcha:
suggested by pgrunewald

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,10 @@ Bug fixes:
   user.get_roles or group.get_roles. Fixes #354
   [pbauer]
 
+- Fix title wrongly set by ``api.content.create`` when called from GS setup
+  handler https://github.com/plone/plone.api/issues/99
+  [gotcha, pgrunewald]
+
 1.6 (2017-02-15)
 ----------------
 
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -106,7 +106,7 @@ def create(
     if IBaseObject.providedBy(content):
         # Will finish Archetypes content item creation process,
         # rename-after-creation and such
-        content.processForm()
+        content.processForm(values=kwargs)
 
     if not id or (safe_id and id):
         # Create a new id from title


gotcha:
http://jenkins.plone.org/job/pull-request-4.3/378 fails with error message related to Jenkins infrastructure.

mauritsvanrees:
Jenkins on 4.3 says: "mr.developer: The package 'plone.api' from auto-checkout has no source information."
`plone.api` is not in `sources.cfg` in coredev 4.3, so it can't be tested with Jenkins on 4.3. On 5.0 and 5.1 you should not have this problem.

Travis fails with:
- 4.3 and 5.0: plone.app.linkintegrity not found. Weird.
- 4.2: ImportError: No module named layout.navigation.root
I get the feeling that the `Makefile` does not work as expected.

gforcada:
@gotcha would you mind rebasing your branch on top of current master? travis was fixed and improved

jensens:
Hence there was no activity for a long time, I propose to close this PR within next two weeks. If you do not feel ok with this, please speak up and provide us a rough schedule.

gotcha:
Am I wrong to hope it runs with 4.3 ?

adamcheasley:
Afaict, the base buildout still extends plone 4.3:
https://github.com/plone/plone.api/blob/master/buildout.cfg#L3
So I'm assuming it should run on 4.3.

adamcheasley:
Looks like the issue might be with jenkins though:
`/tmp/hudson666004923752292000.sh: line 17: bin/jenkins-alltests: No such file or directory`

mauritsvanrees:
See [my comment above](https://github.com/plone/plone.api/pull/319#issuecomment-243785336), from last year. `plone.api` is not used anywhere in coredev 4.3, so Jenkins fails because it cannot find a source. So the Jenkins 4.3 failure can be ignored.

There *is* a `.travis.yml` though, and both the 4.3 and 5.0 tests [pass there](https://travis-ci.org/plone/plone.api/builds/204111314).

So all relevant tests are fine.

But are you sure all changes here are wanted? I see 78 commits and changes in 31 files.  Maybe you can rebase?  (Carefully...)

gotcha:
I thought I had already rebased. Not sure what I did wrong.

Only 2 commits this time. :-)

mauritsvanrees:
Previously you did a merge, which is an entirely different beast. I think it should have the same code as result, but a very different history. Maybe github was confused.

Anyway, much saner and smaller now, thanks.
👍 from me, ready for merge as soon as Jenkins is happy.

PR #321, opened 2016-10-19
Fix an AttributeError in `api.user.revoke_roles`

ale-rt:
`actual_roles` should be a list, but the `get_roles` function might return a tuple.
If that happens you will have an attribute error when calling `actual_roles.remove`.

`get_roles` returns a tuple when it passes here:

- https://github.com/plone/plone.api/blob/3af258a0317d17ed251cb4658f0b034b2e205b3c/src/plone/api/user.py#L243

given that the function `get_local_roles_for_userid` from `AccessControl` always returns a tuple:

- https://github.com/zopefoundation/AccessControl/blob/501afff5636bc6d695b8a2a75e4ff93d3affcac4/src/AccessControl/rolemanager.py#L327

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 ------------------
 
 Fixes:
+- Fix an AttributeError in `api.user.revoke_roles`
+  [ale-rt]
 
 - Typo in the documentation.
   [ale-rt]
--- a/src/plone/api/user.py
+++ b/src/plone/api/user.py
@@ -408,7 +408,8 @@ def revoke_roles(username=None, user=None, obj=None, roles=None):
     if obj is not None:
         # if obj, get only a list of local roles, without inherited ones
         inherit = False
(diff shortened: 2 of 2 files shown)


ale-rt:
@gforcada can this be merged?
The testing failures are not related to this PR.

gforcada:
@ale-rt I'm always wary on merging pull requests with broken tests... let's start travis again and see what's happening

ale-rt:
It seems a problem related with the code-analysis script.

gforcada:
@ale-rt master is already broken, let's first fix master and then we can merge yours, I know that's cumbersome, but master should always be green to begin with, so if we fix it now later pull requests will be easy peasy, if not all of them will be like this one "it's not my fault, let's merge anyway" :-/

ale-rt:
Yeah, I know there is a lot to do to get master green, that's why I started fixing it with #322.
Thanks anyway :)

ale-rt:
Thank you very much @gforcada!

PR #322, opened 2016-10-19
Pep8

ale-rt:


gforcada:
@ale-rt I think that the reason that we did not fix these errors was due to backwards compatibility if I recall correctly... but probably by now we can expect to support 4.3.x and 5.x versions only

ale-rt:
Cool! Be aware that this is a work in progress. 
I was trying to fix the code analysis failures I see in #321 to get it merged.

The big part will be decreasing the complexity of the functions.

adamcheasley:
Yes, sorry, I think two of those are part of the `transition` function which I wrote a little while back. I guess they just need to be broken down a bit.

ale-rt:
@adamcheasley no need to be sorry: that code is great!

ale-rt:
@gforcada:about the compatibility issues...
As far as I can see we are testing plone.api only for Plone >= 4.3.
This PR is going to touch only test files.
I do not see big problems in merging this one (which decreases the code analysis errors).

gforcada:
@ale-rt you are totally right, let's merge this one as soon as you add a change log entry and jenkins is happy about it :-)

ale-rt:
Thanks! Build is running.
Now we should just have problems with the cyclomatic complexity, which will be fun to work with :)

I still have investigate the isort thing (imports look sorted case insensitive, isn't this the new recommended way?).

Issue #290, opened 2015-12-08
plone.api.group.get_groups() error for anonymous user


zopyx:
```
  groups = plone.api.user.get_current().getGroups()
```

raises an "internal" error for Anonymous User

```
  Module plone.api.group, line 2, in get_groups
  Module plone.api.validation, line 100, in wrapped
  Module plone.api.group, line 91, in get_groups
  Module Products.PlonePAS.tools.groups, line 318, in getGroupsForPrincipal
  Module Products.PlonePAS.utils, line 176, in getGroupsForPrincipal
  Module Products.PluggableAuthService.plugins.RecursiveGroupsPlugin, line 87, in getGroupsForPrincipal
  Module Products.PlonePAS.pas, line 189, in getGroups
  Module Products.CMFCore.utils, line 10, in check_getToolByName
  Module Products.CMFCore.utils, line 120, in getToolByName
AttributeError: portal_groups
```

This AttributeError should be better handled.

adamcheasley:
Why not use the safer (although longer):
`groups = api.group.get_groups(user=api.user.get_current())`

jaroel:
Closing this, as getGroups is a PAS method, which doesn't do that much error checking.

    ipdb> user
    <SpecialUser 'Anonymous User'>
    ipdb> user.getGroups()
    *** AttributeError: portal_groups
    ipdb> user.getGroups
    <bound method PluggableAuthService.getGroups of <PluggableAuthService at /acl_users>>
    ipdb> user.getGroups.im_func.func_code
    <code object getGroups at 0x104fd26b0, file "[...]/Products/PlonePAS/pas.py", line 205>

    ipdb> user.aq_base.getGroups
    *** AttributeError: getGroups

zopyx:
Are you serious?

jaroel:
Ah, I misread this completely.
This is a plone.api issue, thanks!

    ipdb> user
    <SpecialUser 'Anonymous User'>
    ipdb> api.group.get_groups(user=api.user.get_current())
    *** AttributeError: portal_groups

jaroel:
Testcase
    ipdb> from AccessControl.users import nobody
    ipdb> plone.api.group.get_groups(user=nobody)
    *** AttributeError: getGroups

jaroel:
Closed by #328

PR #328, opened 2016-10-31
Fix 290 anonymous get groups

jaroel:


jaroel:
Fixes #290

gforcada:
@jaroel would it be possible to merge the commits so that is one commit for the env.py and related changes and another one for the group.py and related changes? :cake:

jaroel:
@gforcada sure, but those are already handled by #327

Issue #313, opened 2016-03-11
api.content.move should warn if you are moving to the parent folder of the element


gforcada:
So if **both** source and target are specified and the parent's source is already target, there's no need to move anything.

This still allows to not pass a target and make renaming happen.

Opinions?

hvelarde:
skipping the operation should be enough, IMO.

gforcada:
Fine with me as well

mauritsvanrees:
Skipping seems fine. I had this recently where running an upgrade step for a second time would try to move an item to the folder where it already was. It worked fine, but was unnecessary.

jaroel:
It is probably replacing the `if target:` check with something like `if target and target is not source.aq_parent`

PR #331, opened 2016-11-05
Avoid a copy & paste on api.conent.move

gforcada:
If source parent's is the target.

Fixes
https://github.com/plone/plone.api/issues/313

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -25,6 +25,9 @@ Fixes:
 - Fix travis and coveralls.
   [gforcada]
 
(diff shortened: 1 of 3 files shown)


jaroel:
@gforcada I've adjusted the test. I've checked this by running the tests without the changes to content.py (FAIL) and with the changes (PASS).

I'll merge this if this makes sense to you too.

gforcada:
@jaroel yes, thanks I was planning to do that as well, just didn't have the time/energy at that moment :-)

PR #332, opened 2016-11-08
Fix translation related tests to use the ``plonelocales`` domain instead `…

thet:
…`passwordresettool``.

Products.PasswordResetTool was removed in Plone 5.1.


thet:
Tests are green, got positive review from @jensens - merge myself.

PR #334, opened 2016-11-10
Registry record default

### LINKED ISSUE #317: Return a default value if no registry record is available
I have seen a common pattern arise in my code:

``` python
record = ISocialLikeSettings.__identifier__ + '.enabled_portal_types'
try:
    enabled_portal_types = api.portal.get_registry_record(record)
except InvalidParameterError:
    enabled_portal_types = []
```

can we have a default value instead?

``` python
record = ISocialLikeSettings.__identifier__ + '.enabled_portal_types'
enabled_portal_types = api.portal.get_registry_record(record, [])
```

ale-rt:
closes #317

### DIFF
--- a/src/plone/api/portal.py
+++ b/src/plone/api/portal.py
@@ -2,8 +2,8 @@
 """Module that provides various utility methods on the portal level."""
 
 from Acquisition import aq_inner
-from datetime import date
 from datetime import datetime as dtime
+from datetime import date
 from email.utils import formataddr
 from email.utils import parseaddr
 from logging import getLogger
@@ -44,6 +44,8 @@ else:
     # file
     import Globals
     PRINTINGMAILHOST_ENABLED = Globals.DevelopmentMode
+
+MISSING = object()
 
 
 def get():
@@ -251,7 +253,7 @@ def show_message(message=None, request=None, type='info'):
 
 
 @required_parameters('name')
-def get_registry_record(name=None, interface=None):
+def get_registry_record(name=None, interface=None, default=MISSING):
     """Get a record value from ``plone.app.registry``
 
     :param name: [required] Name
@@ -259,6 +261,8 @@ def get_registry_record(name=None, interface=None):
     :param interface: interface whose attributes are plone.app.registry
         settings
     :type interface: zope.interface.Interface
+    :param default: The value returned if the record is not found
+    :type default: anything
     :returns: Registry record value
     :rtype: plone.app.registry registry record
     :Example: :ref:`portal_get_registry_record_example`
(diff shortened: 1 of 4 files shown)


jaroel:
Hi Alessandro!

We're using the name `default` instead in `plone.api.content.get_state   https://github.com/plone/plone.api/blob/master/src/plone/api/content.py#L339

This looks like the same kind of thing, right? Can we use `default` in here too?

-Roel

ale-rt:
@jaroel it is fine for me, I am going to update this PR!

src/plone/api/portal.py
```diff
@@ -277,36 +281,56 @@ def get_registry_record(name=None, interface=None):
...
 
-    return registry[name]
+    if default is MISSING:
+        raise InvalidParameterError(msg)
+
+    msg = '{0}\nReturning default {1}.'.format(
+        msg,
+        repr(default)
+    )
+    logger.warning(msg)
```
jaroel: This looks a bit much actually :)
I think I understand why you'd want a warning, just not by default.

gforcada:
@ale-rt would you mind double checking if https://github.com/plone/plone.api/blob/master/docs/portal.rst needs to be updated as well?

ale-rt:
I think the 5.1 job failed for some jenkins issues

src/plone/api/portal.py
jaroel: @ale-rt I'm serious; please don't.

ale-rt: Sorry @jaroel, I think I misunderstood your sentence.
Would you prefer to change the debug level to debug or to use warning.warn so that the log message will appear only once?

ale-rt: Or/and would you change the logged message?

gforcada:
@ale-rt if you rebase on top of current master it will be fine, the 5.1 job failed on a plone.api test that I already fixed on #335

ale-rt:
Thanks @gforcada I am waiting on clarifications by @jaroel about how the logging should happen.
I also agree with him that it is now too much verbose!

src/plone/api/portal.py
jaroel: I'd like no message at all. The user has explicitly passed in a default value, so there's no added value afaic

jensens:
folks, use labels!

ale-rt:
Sorry @jensen!
@gforcada I rebased and now its green.
@jaroel logging was removed, would you mind to review again?

hvelarde:
awesome! thank you very much for this.

jaroel:
Awesome!

PR #341, opened 2017-01-04
339 get roles anonymous

jaroel:


src/plone/api/user.py
```diff
@@ -232,7 +232,11 @@ def get_roles(username=None, user=None, obj=None, inherit=True):
         else:
             username = user.getId()
 
-    user = portal_membership.getMemberById(username)
+    if username is None and user is None:
+        user = portal_membership.getAuthenticatedMember()
+    else:
+        user = portal_membership.getMemberById(username)
```
adamcheasley: Just re-reading the flow here. Not something you've introduced, but it seems that if you pass in the user object, and omit the username, this function will look the user object up again, via the username it gets off the user object. Is that correct?

jaroel: Correct. I tried refactoring this, but then somehow tests for get_roles_in_context start failing.

PR #345, opened 2017-01-24
Allow to only get local roles for groups and do not reassign global r…

pbauer:
…oles as local roles

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -10,7 +10,7 @@ Breaking changes:
 
 New features:
 
-- Allow to only get local roles for groups by passing inherit=False to groups.get_roles()
+- Passing inherit=False to groups.get_roles() will only get local roles for the group.
   [pbauer]
 
 Bug fixes:
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -10,7 +10,8 @@ Breaking changes:
 
 New features:
 
-- *add item here*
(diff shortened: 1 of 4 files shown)


docs/CHANGES.rst
```diff
@@ -10,13 +10,17 @@ Breaking changes:
 
 New features:
 
-- *add item here*
+- Allow to only get local roles for groups by passing inherit=False to groups.get_roles()
```
adamcheasley: I would rephrase this as follows:
"Passing inherit=False to groups.get_roles() will only get local roles for the group."

src/plone/api/group.py
```diff
@@ -219,12 +219,23 @@ def get_roles(groupname=None, group=None, obj=None):
...
+    if obj is None:
+        return group.getRoles()
+    elif inherit:
+        # when context obj is available we bypass getRolesInContext method
+        # from PloneGroup class to use PloneUser class implementation because
+        # PloneGroup class disables all local roles support
+        # see: Products.PlonePAS.plugins.group.PloneGroup
+        roles = super(group.__class__, group).getRolesInContext(obj)
+        return list(roles)
+    elif obj and not inherit:
```
adamcheasley: Could this not just be `else`?

pbauer: Done

docs/CHANGES.rst
pbauer: Done

PR #347, opened 2017-01-31
stop jenkins complaining about "bad quotes"

adamcheasley:


adamcheasley:
I've now added a changelog entry. How do I get that verifier to run again?

docs/CHANGES.rst
```diff
@@ -19,6 +19,9 @@ Bug fixes:
 
 - Fix imports from Globals that was removed in Zope4
   [pbauer]
+  
+- Fix "bad" quotes.
```
hvelarde: you have bad quotes here also ;-)

adamcheasley: Ha! Fair enough.

PR #348, opened 2017-02-01
Fix env tests in Zope4

MatthewWilkes:
No longer try to add Member and Manager roles in role tests, as in Zope4 adding a duplicate role raises an exception.


MatthewWilkes:
Potential bogus failures on 5.1?

PR #352, opened 2017-02-10
Improve: Docs, follow style-guide

svx:
- Improve reST of docs
- Improve line length in /docs
- Move LICENSE into /root of repo => follow style-guide


svx:
@gforcada this makes no sense for me :)

I worked on the docs, which are written in reST, getting CI messages with pep8 seems a bit weird here :)

gforcada:
@svx it's not running the qa tools on the diff, but rather the whole project sorry.

gforcada:
@svx at least https://github.com/plone/plone.api/blob/master/setup.py#L13 needs to be updated (and the next line as well if you moved the LICENSE.txt as well

PR #353, opened 2017-02-10
Do not reassign dynamic roles as local when using group.grant_roles()

### LINKED ISSUE #351: grant_roles() copies dynamic local roles to persistent local roles
`grant_roles(username='foo', roles=['bar'], obj=obj)` first gets the list of current local roles for group foo on obj, then adds the new role and adds the sum as local roles. The problem is that in the first step this also gets roles granted by a dynamic ILocalRoleProvider, thus turning a dynamic role into a persitent one. This can result in permanent access for users who should no longer have access. 

I'm not 100% sure how to fix that. I think the right way is to use `getattr(obj, '__ac_local_roles__', {})` to only get persistent local roles for that object and skip any kind of acquisition and dynamic lookup.

pbauer:
Fixes #351

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,9 @@ Bug fixes:
 - reST syntax in documentation, follow style-guide, adjust setup.py
   [svx]
 
+- Do not reassign dynamic roles as local roles when using group.grant_roles().
+  [pbauer]
+
 1.5.1 (2016-12-06)
 ------------------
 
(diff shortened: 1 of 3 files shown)


ale-rt:
LGTM, I think we need to do something similar for `api.user.grant_roles`:

- https://github.com/plone/plone.api/blob/f25a2f3821a9917aa267066e8f1bf9cd393e977d/src/plone/api/user.py

Issue #354, opened 2017-02-20
get_roles inconsistent 


pbauer:
For users: `get_roles('foo', obj=bar, inherit=False)` uses `obj.get_local_roles_for_userid(username)` which ignores LocalRoles-Adapters.

For groups it uses `lrmanager.getRolesInContext(plone_user, obj)` which includes local_roles set on parents.

Both are wrong should instead return roles that are set only on this object including those provided by LocalRoles-Adapters or from groups that the user is a member of but ignore global roles and roles from parents.

I'm working on a pull-request.

pbauer:
This solution in https://github.com/plone/plone.api/pull/355 is not really beautiful but seems to cover all the use-cases:

```python
plone_user = user.getUser()
principal_ids = list(plone_user.getGroups())
principal_ids.insert(0, plone_user.getId())
roles = set([])
pas = portal.get_tool('acl_users')
for _, lrmanager in pas.plugins.listPlugins(ILocalRolesPlugin):
    for adapter in lrmanager._getAdapters(obj):
        for pid in principal_ids:
            roles.update(adapter.getRoles(pid))
return list(roles)
```

It includes basic local_roles (`obj.__ac_local_roles__`) for the user and all groups that he is a member of. It also handles LocalRoles-Adapters. For groups the only difference is the first line: `plone_user = super(group.__class__, group)`.

thet:
Fixed by #355.

Issue #346, opened 2017-01-25
user.get_roles does not include local roles inherited from groups when using "inherit=False"


pbauer:
As far as I know the `inherit=False` in `user.get_roles(user=foo, obj=bar, inherit=False)` is meant to ignore global roles. What it also does is ignore roles gotten from local roles of groups on the same object. That is inconsistent with `user.get_roles(user=foo, obj=bar, inherit=True)` which returns global and local roles owned by the user and local and gloabl roles inherited from groups the user is in.
The feature was added by @cillianderoiste in 654f749ff7a79eb7a21915b4a1b88f4870c51cd7.

thet:
Fixed by #355

PR #355, opened 2017-02-20
Fix inconsistencies when getting and granting local roles

### LINKED ISSUE #351: grant_roles() copies dynamic local roles to persistent local roles

pbauer:
Fixes #351 for users, #346 and #354

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,8 +14,17 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Do not reassign dynamic roles as local roles when using user.grant_roles().
+  Fixes same issue as #351 for groups.
+  [pbauer]
 
+- Include local roles granted from being in a group when using "inherit=False"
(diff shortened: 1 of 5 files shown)


pbauer:
No, they both return lists now. The set in tests is only for comparisons.

adamcheasley:
Right ok. Fair enough.

jensens:
Good catch!

PR #365, opened 2017-04-04
Raise InvalidParameterError on duplicate id

### LINKED ISSUE #340: plone.api.content.copy fails if id is unchanged
```plone.api.content.copy(foo.myid, bah, 'myid')```
will fail with

```  The id "myid" is invalid - it is already in use.```
Which is unexpected.
You have to omit the id param to make this work

### PR DESCRIPTION
**Pull request by jaroel**
but not when safe_id=True.
Fixes #340

### DIFF
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -262,7 +262,8 @@ def copy(source=None, target=None, id=None, safe_id=False):
     new_id = copy_info[0]['new_id']
     if id:
         if not safe_id and id in target:
-            raise InvalidParameterError
+            msg = "Duplicate ID '{0}' in '{1}' for '{2}'"
+            raise InvalidParameterError(msg.format(id, target, source))
         else:
             return rename(obj=target[new_id], new_id=id, safe_id=safe_id)
     else:
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -261,7 +261,10 @@ def copy(source=None, target=None, id=None, safe_id=False):
 
     new_id = copy_info[0]['new_id']
     if id:
-        return rename(obj=target[new_id], new_id=id, safe_id=safe_id)
+        if not safe_id and id in target:
+            raise InvalidParameterError
+        else:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/plone/api/content.py [hunk 1]
```diff
@@ -261,7 +261,10 @@ def copy(source=None, target=None, id=None, safe_id=False):
 
     new_id = copy_info[0]['new_id']
     if id:
```
jensens: I always get gray hair if someone uses Python built-ins as variables, but that's not your fault here.

src/plone/api/content.py
```diff
@@ -261,7 +261,10 @@ def copy(source=None, target=None, id=None, safe_id=False):
...
-        return rename(obj=target[new_id], new_id=id, safe_id=safe_id)
+        if not safe_id and id in target:
+            raise InvalidParameterError
```
jensens: Do we usually raise errors without any message in plone.api? If so this is ok and we should change it later.

adamcheasley: I'd prefer we moved to raising errors with messages now.

src/plone/api/content.py
jaroel: How about "Duplicate ID '{id}' in '{target}' for '{source}'"?
ie: `InvalidParameterError: Duplicate ID 'contact' in '<ATFolder at about>' for '<ATDocument at contact>'`

src/plone/api/content.py
jaroel: How about "Duplicate ID '{id}' in '{target}' for '{source}'"?
ie: `InvalidParameterError: Duplicate ID 'contact' in '<ATFolder at about>' for '<ATDocument at contact>'`

src/plone/api/content.py
jaroel: It is one of the function arguments, so no cigar!

gforcada: It can always be renamed with a deprecation notice...

jensens: Anyway, this is out of scope. I'll open a new issue.

PR #369, opened 2017-04-14
implement read-only check

loechel:
Implement a ZODB check to see if DB is in read-only Mode.

### DIFF
--- a/docs/env.rst
+++ b/docs/env.rst
@@ -102,6 +102,22 @@ To know if your Plone instance is running in a test runner, use :meth:`api.env.t
         pass  # do something
 
 
+.. _env_read_only_mode_example:
+
+Read-Only mode
+==============
+
+To know if your Zope / Plone instance is running on a read-only ZODB connection use :meth:`api.env.read_only_mode`.
+
+.. code-block:: python
+
+    from plone import api
+
+    in_test_mode = api.env.read_only_mode()
+    if in_test_mode:
+        pass  # do something
+
+
 .. _env_plone_version_example:
 
 Plone version
--- a/docs/env.rst
+++ b/docs/env.rst
@@ -109,12 +109,17 @@ Read-Only mode
 
 To know if your Zope / Plone instance is running on a read-only ZODB connection use :meth:`api.env.read_only_mode`.
 
+**Use-Case:**
+If you run a ZRS or RelStorage cluster with active replication where all replicas are read-only be default.
+You could check if your instance is connected to a read only ZODB or a writeable ZODB.
+Therefore you could adjust the UI to prevent create, delete or update pages are shown.
+
 .. code-block:: python
 
     from plone import api
 
-    in_test_mode = api.env.read_only_mode()
-    if in_test_mode:
+    is_read_only = api.env.read_only_mode()
+    if is_read_only:
         pass  # do something
 
 
--- a/src/plone/api/env.py
+++ b/src/plone/api/env.py
@@ -13,6 +13,7 @@ from plone.api.validation import mutually_exclusive_parameters
 from plone.api.validation import required_parameters
 from zope.globalrequest import getRequest
 
+import Globals
 import traceback
 
 
@@ -206,9 +207,12 @@ def read_only_mode():
 
     :Example: :ref:`env_read_only_mode_example`
     """
-    conn = Globals.DB.open()
-    isReadOnly = conn.isReadOnly()
-    conn.close()
+    isReadOnly = True
+    try:
+        conn = Globals.DB.open()
+        isReadOnly = conn.isReadOnly()
+    finally:
+        conn.close()
     return isReadOnly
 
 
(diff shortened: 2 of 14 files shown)


docs/env.rst
```diff
@@ -102,6 +102,22 @@ To know if your Plone instance is running in a test runner, use :meth:`api.env.t
         pass  # do something
 
 
+.. _env_read_only_mode_example:
+
+Read-Only mode
+==============
+
+To know if your Zope / Plone instance is running on a read-only ZODB connection use :meth:`api.env.read_only_mode`.
```
zupo: would be nice to know the use case when you would need this

loechel: Hi @zupo, my use case is on ZRS or RelStorage, when you have an active replication, where the Replicas are read only and the master goes down.

I do use it to allow my Plone instance to run and disable create, delete and update methods in the UI. One Example: https://github.com/lmu/lmu.contenttypes.blog/blob/master/src/lmu/contenttypes/blog/browser/templates/listing_view.pt#L55

I was asked to contribute this in plone.api, after showing that at a sprint.

src/plone/api/env.py
```diff
@@ -201,6 +202,17 @@ def test_mode():
     return IS_TEST
 
 
+def read_only_mode():
+    """Returns True if you are running the zope instance on an read only ZODB.
+
+    :Example: :ref:`env_read_only_mode_example`
+    """
+    conn = Globals.DB.open()
```
jensens: Do we need to open a new db connection here? Or is it probably better to get the current connection or one from the pool if there is no current one?

docs/env.rst
zupo: What I meant was, add it to the docs :)

src/plone/api/env.py
loechel: You will need just a connection, it does not matter if it is a new or a shared one.

If you have code for getting the current or one from the pool, your are welcome to improve it.

docs/env.rst
loechel: Added.

src/plone/api/env.py
jensens: Ok, if it just open and close one it should be fine. I just wonder if this could be determined once, w/o need to do it every access. Afaik the r/o configuration cannot be changed at runtime.

gforcada: Should this be wrapped in a try/finally? So the connection is always closed?

gforcada:
@loechel after going through all the feedback, would you mind squashing together the commits? There's no need to keep 5 commits for adding such a small feature.

src/plone/api/env.py
loechel: That is the problem, the r/o configuration could change for a ZEO-Client at every request. See ZRS or RelStorage options in plone.recipe.zopeinstance:

    zeo-address = ${ZRS-Master} ${ZRS-Replica1} ... ${ZRS-ReplicaN} 

All ZRS-Replica are read-only by default, but ZRS-Master is writeable. 
ZRS-Master goes down, e.g. OS reboot, and ZRS-Replica takes over, at this moment the runtime config switches from read_only_mode = False to read_only_mode = True and back.

loechel:
@gforcada sure I don't won't to blow the vcs history by such a small feature. I either could squash it in this branch or, the person who merge just do a "squash and merge" Merge.

bildschirmfoto 2017-04-15 um 10 47 41

src/plone/api/env.py
loechel: sure I could wrap it in a try/finally.

gforcada:
@loechel call me *picky*  but I don't like github squash and merge: it has an upside (all commits merged whenever you want that) but it has a really big downside for my personal taste, it does not create a merge commit, which help you know where the merged commit comes from.

Without the merge commit one just sees an stream of commits without any information regarding if they were small changes or pull requests/big features.

Still, better that than having, not one but 5 commits around :+1:

src/plone/api/tests/test_env.py
```diff
@@ -402,6 +402,12 @@ def test_test_mode(self):
         from plone.api.env import test_mode
         self.assertEqual(test_mode(), True)
 
+    def test_read_only_mode(self):
+        """Test that read_only_mode() returns False
+        as we have a write enabled ZODB."""
+        from plone.api.env import read_only_mode
+        self.assertEqual(read_only_mode(), False)
```
hvelarde: use `self.assertFalse()` instead.

loechel:
I have it squashed manually, I hope it is fine now. @gforcada will you have a look and merge if ok.

loechel:
and again. now done a lot to improve this pull and squashed

hvelarde:
please don't mix things: tox has nothing to do with the original scope of this PR; you have to do that on a different PR, after your PLIP has been approved.

loechel:
superseeded by #375 and #376 

I have split them into two pull requests as @hvelarde requested that.

PR #370, opened 2017-05-09
Add disable/enable acquire_local_roles methods to api.content

MrTango:


gforcada:
@MrTango I'm slightly towards a -1 on merging this.

I personally have never needed this nor I see it fitting on the 20% of most used API calls that developers use 80% of the time. Probably the function names don't help much here either, as nearly all plone.api functions are mostly a single word (find, create, get...) or are more specific but broadly used (get/set_registry_record, ...).

Do we want plone.api to be a one stop shop for all? Which criteria should we follow here? Opinions?

MrTango:
Thx @gforcada for your feedback.
The reason why I added this, is that I can grant and revoke already local roles, what I needed, but not do the change for the acquisition flag which is in the same form when you are using the UI to do the task.
I though we decided to fulfill the API more, not only to 20% of the tasks. Also to make it easier later to migrate or exchange things without breaking code of people. In this special case, it took me quite a while to find out the way to set this flag, because of the bad original names and the obscure parameter status.

If you have a better idea for the method names which still make clear what they do, I'm happy to hear it.

jensens:
@gforcada if this is not your 80% (where is this rule coming from?) you don't work much with closed down Plone sites, right? For me, this would be very useful and I am +1 to include this.

sverbois:
@jensens where is this rule coming from?
on the page https://pypi.python.org/pypi/plone.api  we can read "to cover 20% of the tasks any Plone developer does 80% of the time"

jensens:
@sverbois which does not answer where the rule comes from, just where it was written down afterward. 
Anyway, also its difficult to tell what a Plone developers tasks are. For me, I at least find a bunch of functions in plone.api I never use, OTHO some I would like to see are missing for my 20%. So it's primary a POV  unless one would ask >100 Plone developers of different areas.

sverbois:
@jensens I'm for plone.api covers 80% of the tasks any plone developer does :-)

gforcada:
I'm glad other chipped-in 👍 thanks!

Looking at it I  think I get what bothers me: ``enable/disable_acquire_xxxx`` is two verbs together which does not read that natural, nor doesn't convey (to me) the purpose of the functions.

What about ``enable/disable_local_roles_acquisition`` ? Or even without the ``_local`` in it, or it will foul you that you know what is is about?

@jensens as for the 80% rule, one of the concerns (if I remember right, I've never been on the sprints that started it) was that if we bloat plone.api, then it will get loose its purpose:  to be a tiny helper utility.

It can, and should, grow, that's out of question. I was raising my voice, that as I personally have never seen it I would not feel comfortable with it. If it's being used and needed for others, then, sure, it can go in once we find suitable names.

As @MrTango already pointed out, for him it was difficult to find where it was implemented, how to call it, etc etc, the current names (no offense) weren't clear to me at all. That's what doubled my concern of being niche + not discoverable.

So what's actually that flag doing? Preventing acquired roles to go further down? i.e.

- I have folders ``/A`` and inside it ``/A/B``
- I grant ``Editor`` role to user Joe on folder ``/A``, now Joe can edit both content in ``/A`` as well as ``/A/B``
- I flag this property on ``/A/B`` folder, now Joe can **not** edit in content inside /A/B ?

But another user, Magdalena, that has a global ``Editor`` role is unaffected by this and can edit content in both ``/A`` and ``/A/B`` despite that setting?

Is that documented somewhere in docs.plone.org ?

hvelarde:
are there other things we can do with local roles in the context of an object? I think it would be more helpful to think this feature in that context.

MrTango:
The original idea of this package was like this (20%). But we discussed this several times in the 2020 discussions. The conclusion was to improve the Plone Python API and the Rest-API to make it easier to change things in the future without breaking to much. For plone.api this also means that it will cover more then just 20% of the stuff developers need. And I would appreciate this, so that we make it easier to develop addons or scripts for Plone.

@gforcada thx for your input, i like your naming suggestions.
My favorite is enable/disable_local_roles_acquisition, but i would be fine without the local part if people thing it's better.
Your description of using local roles is correct and in my opinion on of the strengths of Plone. The usage in the UI is easy, but on the Python level there is this option missing or at least well hidden in the plone_utils.
With this addition we have all what developers/integrators need to manipulate the local roles. You can even use it in Rapido then I guess.

MrTango:
@hvelarde I'm not sure what you mean. This is already a context feature and also implemented in api.content because of it. You will usually use it in combination with api.group.grant_roles() with the context (obj) argument.

Do I miss anything here?

MrTango:
I updated th naming to disable/enable_roles_acquisition that fits better to the naming group.grant_roles().

jensens:
To answer @hvelarde question: The other important/interesting things one can do is to set/get acquired and directly set roles of a given context. 

The case of roles in context for a specific user (which includes global roles) is covered by ``plone.api.user.get_roles``.

If we introduce a function which sets roles, I would expect it also triggers a reindexObjectSecurity (IIRC this is the method) in order to best support the user of the API.

gforcada:
@MrTango one last thing that is worrying me is the discoverability: api.content.disable_roles_acquistion, is related to content, but *only* related. Specially if you look at the user module as well (api.user and api.group) which is full of security related functions.

What if we move all those api.user/group.(roles/permissions) methods to a brand new api.security ?

It would be surprising to first time users that some roles related methods are on user, some on group, and now yet on content as well.

One of the big strengths of Plone is it's security, so maybe it deserves its own module to highlight that?

I would be fine to merge this at it is and right away after open a new issue/pull request moving all these functions around.

Fortunately (thanks to @jensens) we already have the deprecation guidelines 👍

jensens:
@gforcada I agree that the current API is not consistent here. Overall I am +1 for an own ``plone.api.security`` module. But we should discuss this more in depth.

MrTango:
> If we introduce a function which sets roles, I would expect it also triggers a reindexObjectSecurity (IIRC this is the method) in order to best support the user of the API.

Why introduce, we have this method already with user/group.grant_roles()?

gforcada:
> Why introduce, we have this method already with user/group.grant_roles()?

Him, probably not seeing those functions is probably a sign that a api.security should be set in place? 😄

MrTango:
Regarding the new security category, I'm open for this.
But this needs more discussion before starting to move things around.
For now I thing the best place is in api.content because thats where all content related things are.
And even if we change permissions here, it's still content related.
I was thinking before if it would better in user/groups but after discussing this with @agitator and @jensens decided to place it in content. That would the be the first place to look for it in my opinion.
This might change if introduce more categories and refactoring things.

MrTango:
We also should have a look at the REST-API, so that we can harmonize these two APIs.
It would be a bad idea to have the same functionalities in different places and with different naming's.

gforcada:
@MrTango I agree on adding it on content module right now and open the discussion, so let's do that.

CHANGES.rst
```diff
@@ -10,7 +10,8 @@ Breaking changes:
 
 New features:
 
-- *add item here*
+- Add disable_acquire_local_roles and enable_acquire_local_roles to api.content
```
gforcada: @MrTango that needs updating

PR #375, opened 2017-07-20
Code conventions & tox setup

loechel:
split the stuff from #369 into two pull requests.

This is about testability and enforcing code conventions.

### DIFF
--- a/plone-5.1.x.cfg
+++ b/plone-5.1.x.cfg
@@ -1,6 +1,6 @@
 [buildout]
 extends =
-    https://raw.github.com/collective/buildout.plonetest/master/test-5.0.x.cfg
+    https://raw.github.com/collective/buildout.plonetest/master/test-5.1.x.cfg
     buildout.cfg
 
 test-eggs +=
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,6 +3,19 @@ ignore =
     *.cfg
     .travis.yml
 
+[flake8]
+exclude =
+    bootstrap-buildout.py,
+
+ignore =
+    N801,
+    N802,
+    N803,
+    N805,
+    N806,
+    N812,
+    T000,
+    
 [isort]
 force_alphabetical_sort=True
 force_single_line=True
@@ -12,8 +25,11 @@ not_skip=__init__.py
 
 [coverage:run]
 branch = True
-source = src
-omit =
+source = 
+    src
+
+omit = 
+    *.rst
 
 [coverage:report]
 precision = 2
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,3 +2,21 @@
 ignore =
     *.cfg
     .travis.yml
+
+[isort]
+force_alphabetical_sort=True
+force_single_line=True
+lines_after_imports=2
+line_length=200
+not_skip=__init__.py
+
+[coverage:run]
+branch = True
+source = src
+omit =
+
+[coverage:report]
+precision = 2
+
+[coverage:html]
+directory = reports/coverage
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,14 +8,7 @@ exclude =
     bootstrap-buildout.py,
 
 ignore =
-    N801,
-    N802,
-    N803,
-    N805,
-    N806,
-    N812,
-    T000,
-    
+
 [isort]
 force_alphabetical_sort=True
 force_single_line=True
@@ -25,10 +18,10 @@ not_skip=__init__.py
 
 [coverage:run]
 branch = True
-source = 
+source =
     src
 
-omit = 
+omit =
     *.rst
 
 [coverage:report]
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,123 @@
+[tox]
+envlist =
+    py27-Plone-{4.3,5.0,5.1},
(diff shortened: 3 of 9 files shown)


plone-5.0.x.cfg
```diff
@@ -1,4 +1,8 @@
 [buildout]
+extends =
+    https://raw.github.com/collective/buildout.plonetest/master/test-5.0.x.cfg
+    buildout.cfg
+
 test-eggs +=
     plone.app.contenttypes
```
hvelarde: this is needless, plone.app.contenttypes is the default framework in Plone 5.

plone-5.1.x.cfg
```diff
@@ -0,0 +1,10 @@
+[buildout]
+extends =
+    https://raw.github.com/collective/buildout.plonetest/master/test-5.1.x.cfg
+    buildout.cfg
+
+test-eggs +=
+    plone.app.contenttypes
```
hvelarde: this is needless, plone.app.contenttypes is the default framework in Plone 5.

hvelarde:
awesome! now, if you can explain what all of those option do in some place it would be great for the rest of us to understand the new setup.

loechel:
@hvelarde sure I am writing on that see https://github.com/plone/plone_best_practices_discussion that will be moved into the bobtemplates.plone repo at a certain time. For now it is there and could be discussed.

plone-5.0.x.cfg
loechel: yes it is needless, but I did leave it there as it was there before and has noting to do with this pull request.

plone-5.1.x.cfg
loechel: same here, I just adapt the 5.0 file for 5.1

hvelarde:
is there an alternate way to read the document instead of navigating to a repo?

loechel:
@hvelarde http://plone-best-practices-discussion.readthedocs.io/en/latest/

tox.ini
```diff
@@ -0,0 +1,115 @@
...
+    flake8-debugger
+    flake8-deprecated
+    flake8-isort
+    flake8-pep3101
+    flake8-plone-hasattr
+    flake8-polyfill
+    flake8-print
+    flake8-quotes
+    flake8-string-format
+    flake8-todo
```
gforcada: The official list of flake8 plugins that are expected to work without any whitelisting is maintained in plone.recipe.codeanalysis:

https://github.com/plone/plone.recipe.codeanalysis/blob/master/setup.py#L71-L84

setup.cfg
```diff
@@ -2,3 +2,37 @@
...
+    bootstrap-buildout.py,
+
+ignore =
+    N801,
+    N802,
+    N803,
+    N805,
+    N806,
+    N812,
+    T000,
```
gforcada: As part of the other comment: flake8-naming is not part of the default list of flake8 add-ons, so I would remove these.

Eventually we might want to enable it, but for now it will be too much noise.

gforcada:
@loechel great job!

It is unfortunate thought that we are now going to split configuration in 4 files thought:
- **buildout.cfg** to create the executables
- **setup.cfg** to configure tools
- **.travis.yml** to run tests on Travis
- **tox.ini** to orchestrate the tests builds

If all but one (tox.ini) are really small (ignoring setup.cfg here), that would mitigate things. Is there a way to extend tox configurations? i.e. an ``extends = ``-like functionality from buildout? The configuration you just added is quite verbose (115 lines) and the difference between a tox.ini for plone.api and one for plone.app.dexterity will be rather low.

Anyway, life is not perfect, but we should try to see if we can improve this later on.

👍 on merging it as soon as the flake8 list is fixed.

loechel:
@gforcada flake8 now fixed.

About the multiple config files. I knew @svx is working on making whole Plone pip installable, so then buildout.cfg in each package could go away. 

But I guess, we will always have a small overhead on config files in each project root. 

I hope I could work on the bobtemplates.plone extensions for best practice, than, we could compare and check very easily all setup.cfg and tox.ini's and so on in future.

loechel:
@gforcada will you merge?

hvelarde:
@loechel Travis builds are broken; no PR should be merged with builds broken.

loechel:
@hvelarde we should not mix to many things. This branch did not change anything on the travis-ci side so that errors used to be already in the code. But yes I will do a new Pull request after #375 and #376 to fix all tests and code conventions errors.

jensens:
@loechel - I merged both.

Btw. since when do we enforce trailing commas?

loechel:
@jensens Thank you. 
Will start working today on the fixes.

About the trailing commas, I don't know I just added those flake8 plugins @gforcada referenced.

hvelarde:
since today? ;-)

jensens:
I usually add trailing commas in my code, but I am against enforcement.

loechel:
well I think that is something the testing and framework team should discuss. 

The enforcing here is just introduced, by adjusting the set of used flake8 Plugins and removing all exceptions in setup.cfg - This was requested in a review by @gforcada, so probably he knows why flake8 now tries to enforce that and why it is inconsistent with plone.recipe.code-analysis.

PR #376, opened 2017-07-20
implement read-only check on ZODB

loechel:
superseed #369 

to make one Pull request make exactly one thing.

PR #377, opened 2017-07-26
Fix tests

loechel:
as said in #375 and #376 this pull just fixes all open errors.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -33,9 +33,3 @@ imports = True
 pep3101 = False
 debug-statements = True
 return-status-codes = True
-
-[versions]
-plone.api =
-
-setuptools = 33.1.1
-zc.buildout = 2.8.0
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -1,10 +1,11 @@
 [buildout]
 
 package-name = plone.api
-package-extras = [develop, test]
+package-extras = [test]
 test-eggs =
-    Pillow
-    plone.app.contenttypes
+    plone.api
+#    Pillow
+#    plone.app.contenttypes
 
 parts +=
     omelette
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -4,8 +4,6 @@ package-name = plone.api
 package-extras = [test]
 test-eggs =
     plone.api
-#    Pillow
-#    plone.app.contenttypes
 
 parts +=
     omelette
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -8,7 +8,6 @@ test-eggs =
 
 parts +=
     omelette
-    code-analysis
 
 # network speedup
 socket-timeout = 5
@@ -22,14 +21,3 @@ allow-hosts =
 [omelette]
 recipe = collective.recipe.omelette
 eggs = ${test:eggs}
-
-[code-analysis]
-recipe = plone.recipe.codeanalysis
-directory = ${buildout:directory}/src/plone/api
-flake8-max-complexity = 20
-utf8-header = True
-clean-lines = True
-imports = True
-pep3101 = False
-debug-statements = True
-return-status-codes = True
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -50,7 +50,7 @@ def create(
     id=None,
     title=None,
     safe_id=False,
-    **kwargs
+    **kwargs  # NOQA: C816
 ):
     """Create a new content item.
 
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -97,7 +97,7 @@ def create(
             "Cannot add a '{0}' object to the container.\n"
             'Allowed types are:\n'
             '{1}\n'
-            '{2}'.format(type, '\n'.join(sorted(types)), e.message)
+            '{2}'.format(type, '\n'.join(sorted(types)), e.message),
         )
 
     content = container[content_id]
@@ -187,7 +187,7 @@ def move(source=None, target=None, id=None, safe_id=False):
     # If no target is given the object is probably renamed
     if target and source.aq_parent is not target:
             target.manage_pasteObjects(
-                source.aq_parent.manage_cutObjects(source_id)
+                source.aq_parent.manage_cutObjects(source_id),
             )
     else:
         target = source.aq_parent
@@ -256,7 +256,7 @@ def copy(source=None, target=None, id=None, safe_id=False):
         target = source.aq_parent
 
     copy_info = target.manage_pasteObjects(
-        source.aq_parent.manage_copyObjects(source_id)
+        source.aq_parent.manage_copyObjects(source_id),
     )
 
     new_id = copy_info[0]['new_id']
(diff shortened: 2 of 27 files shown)


loechel:
could someone please look into those remaining code conventions reports. I don't want to remove any todos. For the C816 Waring I even don't know how to fix that without an NOQA: C816 comment

@gforcada might it be possible to also add the Jenkins CI Link for 4.3?

loechel:
Sorry but here is something wired, isort and isort-apply did interprete the import rules differently. 

at some moment, isort complains about a specific file (test_portal.py) that the imports a incorrectly sorted, isort-apply sorts them, the file is checked in, and isort complains again, but file is unchanged in git.

Did anybody has an idea?

loechel:
OK, I am more and more confused by plone.api, how ever could that package be stable, the tests are not deterministic.

Second why is elementar test function in doctests, didn't we agree to get ride of doctests?

loechel:
Ok I now have fixed the most obvious test failures. 
There are 3 flake8 errors that we should look on together.

buildout.cfg
```diff
@@ -1,14 +1,14 @@
 [buildout]
 
 package-name = plone.api
-package-extras = [develop, test]
+package-extras = [test]
 test-eggs =
-    Pillow
-    plone.app.contenttypes
+    plone.api
+#    Pillow
```
adamcheasley: Suggest actually removing these lines, rather than just commenting them out.

src/plone/api/group.py
```diff
@@ -17,7 +17,7 @@ def create(
     title=None,
     description=None,
     roles=[],
-    groups=[]
+    groups=[],
```
adamcheasley: Unrelated to this PR, but we really shouldn't be using mutable default args.

buildout.cfg
loechel: sure if all agree on that

src/plone/api/group.py
loechel: we should keep this in mind, and either do this on this pull request or a new following one.

I think this repository did need a close look on conventions and best practices. I do see a lot of things that does look a bit scary.

src/plone/api/content.py
```diff
@@ -187,7 +187,7 @@ def move(source=None, target=None, id=None, safe_id=False):
     # If no target is given the object is probably renamed
     if target and source.aq_parent is not target:
             target.manage_pasteObjects(
-                source.aq_parent.manage_cutObjects(source_id)
+                source.aq_parent.manage_cutObjects(source_id),
```
hvelarde: I tend to agree with @jensens: enforcing trailing commas all over the places makes no sense and is against out own code style guidelines; this should be written instead:

```python
target.manage_pasteObjects(
    source.aq_parent.manage_cutObjects(source_id))
```

src/plone/api/tests/base.py
```diff
@@ -58,6 +58,10 @@ def tearDownZope(self, app):
 
 FIXTURE = PloneApiLayer()
 INTEGRATION_TESTING = IntegrationTesting(
-    bases=(FIXTURE,), name='PloneApiLayer:Integration')
+    bases=(FIXTURE, ),
```
hvelarde: same here:

```python
INTEGRATION_TESTING = IntegrationTesting(
    bases=(FIXTURE,), name='PloneApiLayer:Integration'))
```

src/plone/api/tests/test_group.py
```diff
@@ -480,17 +481,27 @@ def test_grant_roles_in_context(self):
...
+                    groupname='foo',
+                    obj=folder,
+                    inherit=False,
+                ),
+            ),
         )
         self.assertEqual(
             ROLES,
-            set(api.group.get_roles(group=group, obj=folder, inherit=False)),
+            set(api.group.get_roles(group=group, obj=folder, inherit=False,)),
```
hvelarde: this is even worst; seems to me the flake8 plugin is just broken.

jensens: I agree, that not what we want. 

I also consider the plugin flake8-commas as broken. [flake8-commas](https://pypi.python.org/pypi/flake8-commas) states "Now you can avoid those annoying merge conflicts on dictionary and list diffs." But it also takes parameters into account, which is more than it should.

I am +1000 to disable the plugin until it works.

src/plone/api/content.py
jensens: against you example I like the trailing bracket on a new line. but I don't see a comma needed.

src/plone/api/tests/test_group.py
jensens: I filed a bug report https://github.com/flake8-commas/flake8-commas/issues/43

loechel: Ok I could agree on that, but it seams also that the tailoring in plone.recipe.code-analysis does not enforce that, while the plugin itself in flake8 tries that.

I will remove that plugin for the moment from the plugins list in tox and remove stupid commas in the files.

src/plone/api/content.py
gforcada: @hvelarde no please!! 😄 Keep the parenthesis on its own, for two reasons:

- there is a clear clue where the statement finishes (as it will be aligned with the ``t`` of target
- if eventually someone needs to add a second parameter the diff will be just one single line, rather than two

src/plone/api/tests/base.py
gforcada: @hvelarde seems we don't agree on the style, that's a hard no go for me (at least on my projects).

As soon as a statement can not fit into a single line, then everything needs to be on their own, mostly for readability. This guy explains it way better than this short comments of mine: http://treyhunner.com/2017/07/craft-your-python-like-poetry/ and I wholeheartedly agree with him 😄

src/plone/api/tests/test_group.py
gforcada: Sorry for being late on the party (I've been really busy at work).

I could be ok with not enforcing a comma in this specific case, but the statement is *broken* already by having a statement with multiple parameters as a second argument. To sum it up: http://treyhunner.com/2017/07/craft-your-python-like-poetry/

I would change that to:
```
expected_roles = api.group.get_roles(
    group=group,
    obj=folder,
    inherit=False,
)
self.assertEqual(
    ROLES,
    set(expected_roles),
)
```

This is, by far, more readable and has no problem. Readability counts, not cramming as much as you can on a single line of text 😃

loechel:
OK, I would say this is now ready to merge. 
I have reviewed the whole code, reformat for readability and fix all flake8 warning we have had. Also I did wrote some new tests to raise test coverage and fixed several issues with the tests where those haven't been deterministic. 

@jensens @gforcada please review and merge if you feel that this is ok for the moment.
We should have a open space at Plone Conf and talk about Coding Conventions and best Practices in Plone coding.

loechel:
isort still seams not stable, we should have a look at that.

src/plone/api/tests/test_group.py
graingert: `inherit=False,)` should be prohibited not required... I can't repeat this error

PR #387, opened 2017-09-23
Change content.delete to allow obj=None and objects=[] or objects=None.

### LINKED ISSUE #383: Unexpected behavior in content.delete
If I do:
```
api.content.delete(objects=[])
```
I expect that the function will do nothing and exit without errors. However, an exception is raised. 

The reason is the first line of code of the `delete` function:
```
objects = objects or [obj]
```
Since `objects` is an empty list and `obj` is `None` (the default value) we end up with `objects` being `[None]`.

An easy way to fix is to filter out `None` objects from the resulting list:
```
objects = objects or [obj]
objects = [o for o in objects if o is not None]
if not objects:
    return
```

jaroel:
Fixes #383

### DIFF
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -735,7 +735,6 @@ class TestPloneApiContent(unittest.TestCase):
     def test_delete_no_objs(self):
         # Check that we allow passing in an empty list of objects.
         api.content.delete(obj=None, objects=[])
-        api.content.delete(obj=None, objects=None)
 
     def test_delete_ignore_linkintegrity(self):
         """Test deleting a content item with a link pointed at it."""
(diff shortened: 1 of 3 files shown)


src/plone/api/tests/test_content.py
```diff
@@ -732,6 +732,10 @@ def test_delete_multiple(self):
         self.assertNotIn('copy_of_about', container)
         self.assertNotIn('about', container['events'])
 
+    def test_delete_no_objs(self):
+        # Check that we allow passing in an empty list of objects.
+        api.content.delete(obj=None, objects=[])
```
hvelarde: this test is wrong as `obj` and `objects` are mutually exclusive parameters:

    @mutually_exclusive_parameters('obj', 'objects')

mauritsvanrees: The test passes, so apparently it is fine.
I see that the `mutually_exclusive_parameters` decorator uses [`_get_supplied_args`](https://github.com/plone/plone.api/blob/master/src/plone/api/validation.py#L30), which ignores arguments that are `None`. That seems why it is okay.

PR #388, opened 2017-09-24
Don't rename an object when the id already is the target id.

### LINKED ISSUE #361: api.content.move with id tries to rename when it doesn't need to
`api.content.move` accepts an `id` parameter to rename the item that is being moved. It tries to do this even if the item already has that id, and in that case raises an exception. I think it would be better if it skips trying to rename in the case where the item already has the specified id.

jaroel:
Fixes #361

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,9 @@ New features:
 
 Bug fixes:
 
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
Thanks. I have merged it manually to fix an obvious conflict in the changelog.

PR #394, opened 2017-10-22
Python 2 / 3 compatible imports

rudaporto:


src/plone/api/portal.py
```diff
@@ -441,7 +442,7 @@ def translate(msgid, domain='plone', lang=None):
     :param lang: target language
     :type lang: string
     :returns: translated message
-    :rtype: unicode
+    :rtype: six.text_type
```
mauritsvanrees: This is a docstring and it seems weird to have `six.text_type` end up in the [documentation](https://docs.plone.org/develop/plone.api/docs/api/portal.html#plone.api.portal.translate).  But that is documentation for developers, so I guess it is okay.

There is one test failure on Plone 5.1, but that is an unrelated one that has meanwhile been fixed.
Seems good to merge. I will do that.

PR #395, opened 2018-02-09
fix #99 Call processForm with empty values.

### LINKED ISSUE #99: api.content.create replaces archetype-content title with site title during Plone site creation

david-batranu:
From what I can see, widgets mostly do the right thing and return empty_marker if field.getName() is missing from the request. Perfect when programatically adding content through invokeFactory/api.content.create as the set values don't get modified.

### DIFF
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -363,7 +363,7 @@ class TestPloneApiContent(unittest.TestCase):
             container=container,
             type='Folder',
             id='test-folder',
-            title='Test folder'
+            title='Test folder',
         )
         assert folder
         self.assertEqual(folder.id, 'test-folder')
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -345,6 +345,43 @@ class TestPloneApiContent(unittest.TestCase):
         # check that the exception is the one we raised
         self.assertEqual(ude.exception.reason, unicode_exception_message)
 
+    def test_create_at_with_title_in_request(self):
+        """ Test that content gets created with the correct title, even if
+            request.form['title'] already exists and has a different value.
+            This can occur, for example, when adding a Plone with an enabled
+            product that creates a site structure. In that case, the 'title'
+            would be that of the portal.
+            Only AT content types are affected, due to content.processForm.
+        """
+        leaked_title = 'This should not be set on content items'
+        self.layer['request'].form['title'] = leaked_title
+
+        container = self.portal
+
+        # Create a folder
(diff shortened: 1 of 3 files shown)


gotcha:
Would you be less lazy than I was and add a test ?

david-batranu:
@gotcha sure! if tox ever finishes fetching the 9 billion packages Plone depends on. Hopefully this decade.

david-batranu:
@gotcha done! Added a test in `test_content.TestPloneApiContent`, hopefully that's the right place for it. 

In the end I had to give up on the `tox` tests as they kept running buildout and re-fetching eggs on each run. I ended up running the tests on one of my existing buildouts.

This took way more time than I anticipated.

gotcha:
@david-batranu Thanks a lot for the test !
As you can see, Changelog verifier is complaining about a missing entry. Would you mind adding it ?

(I have also triggered the Jenkins build)

src/plone/api/tests/test_content.py
```diff
@@ -345,6 +345,43 @@ def unregister_indexer():
...
+        container = self.portal
+
+        # Create a folder
+        folder = api.content.create(
+            container=container,
+            type='Folder',
+            id='test-folder',
+            title='Test folder'
+        )
+        assert folder
```
hvelarde: just being curious: what's the point on this assertions?

david-batranu: @hvelarde thank you for your feedback! I copied the test from `test_create_content` and forgot those. I'll clean them up right away.

gotcha:
@david-batranu I started the jobs on Plone Jenkins. Do you know how to start them ?

david-batranu:
@gotcha I wanted to start them but it asked for GitHub account linking, and I did not want to do that.

gotcha:
@david-batranu Thanks again

PR #399, opened 2018-03-20
Simplification/minor speedup: user.* permission checks

jensens:
Simplification/minor speedup:
Permissions checks now directly use AccessControl.
Technical its now exact the same as before.
Before a tool lookup was needed, calling a utility function, calling AccessControl.

### DIFF
--- a/src/plone/api/user.py
+++ b/src/plone/api/user.py
@@ -2,6 +2,7 @@
 """Module that provides functionality for user manipulation."""
 
 from AccessControl.Permission import getPermissions
+from AccessControl.SecurityManagement import getSecurityManager
 from contextlib import contextmanager
 from plone.api import env
 from plone.api import portal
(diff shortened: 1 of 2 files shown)


src/plone/api/user.py
```diff
@@ -298,11 +299,10 @@ def get_permissions(username=None, user=None, obj=None):
...
     result = {}
     with context:
-        portal_membership = portal.get_tool('portal_membership')
-        permissions = (permission[0] for permission in getPermissions())
-        for permission in permissions:
+        for record in getPermissions():
+            permission = record[0]
             result[permission] = bool(
-                portal_membership.checkPermission(permission, obj),
+                getSecurityManager().checkPermission(permission, obj),
```
ale-rt: If you are really interested in performance I would define getSecurityManager() out of the for cycle.
```
sm = getSecurityManager()
permissions = (record[0] for record in getPrermissions())
result = {permission: bool(sm.checkPermission(permission, obj) for permission in permissions}
```

jensens: indeed!

src/plone/api/user.py
```diff
@@ -298,13 +299,9 @@ def get_permissions(username=None, user=None, obj=None):
 
     result = {}
```
ale-rt: This line is no longer needed, but this can be cleaned also later

jensens:
@ale-rt I removed the line, I would say ready if tests go green, thanks for review!

ale-rt:
Thanks for the enhancement :)

PR #402, opened 2018-05-25
Python3

pbauer:


src/plone/api/content.py
```diff
@@ -100,7 +100,7 @@ def create(
             '{message}'.format(
                 obj_type=type,
                 allowed_types='\n'.join(sorted(types)),
-                message=e.message,
+                message=e.args[0],
```
hvelarde: AFAIK this should be fixed with `str(e)`; see: https://stackoverflow.com/a/7075416

src/plone/api/group.py
```diff
@@ -95,7 +95,7 @@ def get_groups(username=None, user=None):
             groups = group_tool.getGroupsForPrincipal(user)
         except AttributeError as e:
             # Anonymous users from the Zope acl_users folder will fail on this
-            if 'portal_groups' in e.message:
+            if 'portal_groups' in e.args[0]:
```
hvelarde: AFAIK this should be fixed with `str(e)`; see: https://stackoverflow.com/a/7075416

src/plone/api/tests/test_content.py
```diff
@@ -163,7 +163,7 @@ def test_create_constraints(self):
         # in the InvalidParameterError message
         self.assertIn(
             'No such content type: foo',
-            cm.exception.message,
+            cm.exception.args[0],
```
hvelarde: AFAIK this should be fixed with `str(e)`; see: https://stackoverflow.com/a/7075416

PR #406, opened 2018-08-10
Refactor Travis configuration

hvelarde:
### Motivation
* Running a test matrix in Travis is currently easier than in Jenkins because you don't have to manually run the jobs, and re-run them at every other change
* Using tox on top of Travis is overkill and unnecessary
* Travis CI builds are currently broken
* Coveralls is not working
* Code analysis is not being run
* Include check for Python 3 compatibility
### TODO
* Run documentation testing (I have no experience on that and I need help)
* Configuration clean-up; there are many things that can be removed

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[report]
+include = src/plone/api/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,0 @@
-[report]
-include = src/plone/api/*
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 *.py[cod]
 *~
 /.Python
-/.coverage*
 /.installed.cfg
 /bin/
 /build/
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.py[cod]
 *~
 /.Python
+/.coverage*
 /.installed.cfg
 /bin/
 /build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,47 +1,40 @@
 language: python
+python: 2.7
 sudo: false
 cache:
   pip: true
   directories:
-    - $HOME/buildout-cache
+  - $HOME/.pylint.d
+  - eggs
 
 global:
   - secure: "nTXMNU2aYbCtvKD74PndF8Xm2h6IvsxPjr6vj45AxjXmaB7Wf5oGi/4b6ObV\nexBtUQkFnc+M0ThGyUznOCPi3YxcbTzsyhQYhHUtmtw/6QRyYgo4E0GLDItU\n37Ff6wpxl3NMMJTvpi6SOVUvIJUl3+cs+4bkMkS48RSFGg2rGmo="  # Coveralls.io token
+env:
+  matrix:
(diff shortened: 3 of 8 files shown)


.coveragerc
```diff
@@ -0,0 +1,2 @@
+[report]
```
loechel: please don't 
add such config to setup.cfg

.gitignore
```diff
@@ -2,7 +2,6 @@
 *.py[cod]
 *~
 /.Python
-/.coverage*
```
loechel: why do you remove that? it has nothing to do in version control

loechel:
When you whant a multi Python test setup you should always have tox. 
Even if travis might be cabable of doing that on its own, it makes it more complex.

hvelarde:
@loechel look at your current build, that BTW is not working, and look at what I have done. which one is more complex? if you want to use tox, just use it, but you can force anybody to do so.

.travis.yml
```diff
@@ -1,47 +1,40 @@
...
-        env: TOXENV=py27-plone-4.3
-      - python: "2.7"
-        env: TOXENV=py27-plone-5.0
-      - python: "2.7"
-        env: TOXENV=py27-plone-5.1
-      - python: "2.7"
-        env: TOXENV=isort,flake8,docs,
 
 before_install:
-    - mkdir -p $HOME/buildout-cache/{downloads,eggs,extends}
```
hvelarde: this is what I call a "a test setup from old times"; caching has been around Travis since 2016.

loechel:
@hvelarde ok, I will look more into detail about the current status and do an own proposal how to fix it. 

I was preparing a Python 3 Build for it, but as it is a dependency thing we could not test something if it is not compatible.

.travis.yml
loechel: sure, tell me how to do it better?

if you use buildout and not the standard tools like pip which is known for travis you have to tell them which directory to cache:

https://github.com/plone/plone.api/blob/master/.travis.yml#L3-L6

So the maked line is necessary to be idempotent and deterministic.

jensens:
Best look at bobtemplates plone for best practices. We really tried hard to make it perfect there and I always copy from there even to older projects. 

I would prefer to first add a well discussed and we'll documented multi python setup to bobtemplates plone. This may include tox if it helps solving the problem. With upcoming 5.2 we need it anyway. If we have it perfectionated there, we can start using it everywhere. This would help us to not have a forest of different solutions around, creating headache every time a new stumbles over them.

We can use plone.api as a test drive for this perfectionated and we'll documented solution.

.travis.yml
hvelarde: it isn't; if you check what I did you'll learn how to fix it. we have been using it that way for years and we already had that discussion in bobtemplates.plone as @jensens mentioned.

hvelarde:
as mentioned: I'm not against using tox in local environments, but against using tox in Travis, because it solves no problem and just adds another layer of complexity to the build.

jensens:
There was no activity on this for a while. I close this request for now. If you plan to work further on this please reopen and go on.

Issue #364, opened 2017-03-31
plone.api 1.6.1 doesn't populate the start date field of type Event (Archetype), when we use api.content.create


idgserpro:
I have:

```python
api.content.create(
    container=portal,
    type='Event',
    title=u'My event',
    startDate=DateTime(today), 
    endDate=DateTime(tomorrow),
)
```

So:

```python
>>> portal['my-event'].startDate
None
```
I'm using Plone 4.3.13

hvelarde:
@gotcha this problem seems to be caused by your commit: https://github.com/plone/plone.api/commit/f73bddc570c05f7a1a4cea01c75b1cea6538676e

idgserpro:
This has been fixed. See: #407

Issue #367, opened 2017-04-05
Why we are not testing against Plone 4.3?


hvelarde:
I just discovered that PR are not being tested against Plone 4.3; AFAIK, Plone 4.3 is still supported and the lack of testing is opening the door to bugs like #364.

gforcada:
It is tested (supposedly): https://github.com/plone/plone.api/blob/master/.travis.yml#L9 where is 4.3 missing?

hvelarde:
from Jenkins; you can confirm that on any PR.

mauritsvanrees:
Nothing in 4.3 core is using plone.api, and it is not in the version pins or in the sources. Also, plone.api has only been officially added in Plone 5.0. So it seems normal that it is not tested on Jenkins for 4.3.

Travis tests it with the latest Plone 4.3 release. That should be good enough. Regressions caused by a new Plone release will be caught the next time a Travis job is run. plone.api can then be fixed and a new release made.

But yes, if we wanted to, we could probably test plone.api on Jenkins 4.3 too, to catch regressions caused by changes to core packages before they end up in a release.
Actually: it may be tricky: it can only be done by adding `plone.api` to the test eggs so it ends up in the `bin/test` script. And this means that other core packages could start to depend on `plone.api` without us noticing it, if reviewers aren't careful.
So the best way then would be to create a `bin/test-with-plone-api` script and run the `plone.api` tests with that. It should be doable, but I am not enthousiastic. It would need changes to the Jenkins scripts so this new test script is run. Or `bin/alltests` would somehow need to use the `bin/test` script for all packages except for `plone.api` which would need to be run with the new script.
Given this complexity, I am inclined to let it go.

hvelarde:
thanks for the clarification.

PR #407, opened 2018-08-10
Add test cases for some weird issues with Event content type

hvelarde:
# Test only; do not merge!

* https://github.com/plone/plone.api/issues/364
* https://github.com/plone/plone.app.contenttypes/issues/465

### DIFF
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -376,6 +376,46 @@ class TestPloneApiContent(unittest.TestCase):
 
         self.assertEqual(page.title, 'Test document')
 
+    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
+    def test_create_at_event(self):
+        """https://github.com/plone/plone.api/issues/364"""
+        from DateTime import DateTime
+        today = DateTime()
+        tomorrow = today + 1
+        event = api.content.create(
+            container=self.portal,
+            type='Event',
+            title=u'My event',
+            startDate=today,
+            endDate=tomorrow,
+        )
+        self.assertEqual(event.startDate, today)
+        self.assertEqual(event.endDate, tomorrow)
+        results = api.content.find(Title=u'My event')
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].start, today)
+        self.assertEqual(results[0].end, tomorrow)
+
+    @unittest.skipUnless(HAS_PACONTENTYPES, 'Dexterity only')
+    def test_create_dx_event(self):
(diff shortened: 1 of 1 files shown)


hvelarde:
@idgserpro I tested both weird issues we are having in collective.cover and was not able to reproduce them here; I think we have a problem in our test suite setup and we would need some help.

hvelarde:
@idgserpro any feedback?

idgserpro:
@hvelarde are you sure that the ```plone.api``` tests are run on the coredev of ```Plone 4.3```? It's not in [tests.cfg](https://github.com/plone/buildout.coredev/blob/4.3/tests.cfg).

And the travis buildout is not working.

idgserpro:
@hvelarde see #367

hvelarde:
@idgserpro I tested locally in my own machine and tests are passing; you can do the same.

idgserpro:
@hvelarde I run the tests locally and they really passed.

It looks like #364 has been fixed. About the dx problem I do not know much.

Do these two problems occur in cover?

hvelarde:
yes they do, and that's why I think we have a problem with the test fixture.

src/plone/api/tests/test_content.py
```diff
@@ -376,6 +376,46 @@ def test_create_at_with_title_in_request(self):
 
         self.assertEqual(page.title, 'Test document')
 
+    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
+    def test_create_at_event(self):
```
idgserpro: @hvelarde why do not you add this test to the package? It will prevent #364  from occurring again.

hvelarde: Archetypes is going to die soon.

src/plone/api/tests/test_content.py
```diff
@@ -376,6 +376,46 @@ def test_create_at_with_title_in_request(self):
...
+        self.assertEqual(event.startDate, today)
+        self.assertEqual(event.endDate, tomorrow)
+        results = api.content.find(Title=u'My event')
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].start, today)
+        self.assertEqual(results[0].end, tomorrow)
+
+    @unittest.skipUnless(HAS_PACONTENTYPES, 'Dexterity only')
+    def test_create_dx_event(self):
+        """https://github.com/plone/plone.app.contenttypes/issues/465"""
```
idgserpro: @hvelarde  what's the relationship between plone/plone.app.contenttypes#465 and ```plone.api```. It seems to me more a problem of ```plone.app.contenttype```.  What's purpose of this test?

@gbastien FYI.

PR #410, opened 2018-08-14
Test to confirm error in Archetype Collection creation

idgserpro:
refs. #413

### DIFF
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -376,6 +376,22 @@ class TestPloneApiContent(unittest.TestCase):
 
         self.assertEqual(page.title, 'Test document')
 
+    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
+    def test_create_at_collection(self):
+        """Test create at Collecition."""
+        collection = api.content.create(
+            container=self.portal,
+            type='Collection',
+            title='Mandelbrot set',
+            description='Image gallery of a zoom sequence',
+            query=[{
+                'i': 'Type',
+                'o': 'plone.app.querystring.operation.string.is',
+                'v': ['Image'],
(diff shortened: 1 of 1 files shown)


idgserpro:
@hvelarde note that the error that occurs in creating the collection occurs here as well.

hvelarde:
now we must open an issue describing the problem :)

gbastien:
Hi @idgserpro I did not see this PR... We have the same problem, still need to decide how to fix it...

Gauthier

gbastien:
Hi @idgserpro @hvelarde 

I can confirm that if I change {None: None} to {'':''}, bool({'':''} is still True so it is used by AT's processForm and the test you added pass as well as the test_create_at_with_title_in_request that was added regarding the {None: None}.

So I would go for that, what do you think?

Gauthier

hvelarde:
if tests pass I think it could be fine; just be careful to not reintroduce another issue with Archetypes-based Event content type. I would suggest to add tests on #407 to this pull request as well.

idgserpro:
@gbastien I also think ```{'': ''}``` will work. I agree with this solution. If you could create another PR would be great. I have no contributor agreement. Feel free to add the test of this PR and the test suggested by @hvelarde.

gbastien:
@idgserpro ok I will create a separated PR and keep your test.

@hvelarde so I would add the tests of PR #407 as well to the new PR to make sure it works with AT and DX event?  I totally agree, why PR #407 was not merged?

I will propose a PR tomorrow with all this together.

Thank you for review,

Gauthier

hvelarde:
@gbastien it was not merged because I though it was unnecessary, but @idgserpro was right and now I think is important to avoid regressions.

idgserpro:
Replaced for #414

PR #414, opened 2018-09-06
Fix falining AT Collection creation when using api.content.create.

gbastien:
I also added the AT/DX event creation related tests from PR #407.

With this, the Collection creation is correct.

Thank you,

Gauthier

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,7 +21,7 @@ Bug fixes:
 - fix typos in doc strings
   [tkimnguyen]
 
-- Fix falining AT Collection creation when using api.content.create.
+- Fix failing AT Collection creation when using api.content.create.
   [gbastien]
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,9 @@ Bug fixes:
 - fix typos in doc strings
   [tkimnguyen]
 
+- Fix falining AT Collection creation when using api.content.create.
+  [gbastien]
+
 
 1.8.4 (2018-04-24)
 ------------------
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -376,9 +376,8 @@ class TestPloneApiContent(unittest.TestCase):
 
         self.assertEqual(page.title, 'Test document')
 
-    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
-    def test_create_at_collection(self):
-        """Test create at Collecition."""
+    def test_create_collection(self):
+        """Test create a Collection."""
         collection = api.content.create(
             container=self.portal,
             type='Collection',
--- a/src/plone/api/tests/test_content.py
+++ b/src/plone/api/tests/test_content.py
@@ -376,6 +376,62 @@ class TestPloneApiContent(unittest.TestCase):
 
         self.assertEqual(page.title, 'Test document')
 
+    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
+    def test_create_at_collection(self):
+        """Test create at Collecition."""
+        collection = api.content.create(
+            container=self.portal,
+            type='Collection',
+            title='Mandelbrot set',
+            description='Image gallery of a zoom sequence',
+            query=[{
+                'i': 'Type',
+                'o': 'plone.app.querystring.operation.string.is',
+                'v': ['Image'],
+            }],
+        )
+        self.assertEqual(collection.Title(), 'Mandelbrot set')
+
+    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
+    def test_create_at_event(self):
+        """https://github.com/plone/plone.api/issues/364"""
+        from DateTime import DateTime
+        today = DateTime()
+        tomorrow = today + 1
+        event = api.content.create(
+            container=self.portal,
+            type='Event',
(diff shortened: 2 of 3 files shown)


gbastien:
Hi @idgserpro FYI

src/plone/api/tests/test_content.py
```diff
@@ -376,6 +376,64 @@ def test_create_at_with_title_in_request(self):
...
+        self.assertEqual(event.startDate, today)
+        self.assertEqual(event.endDate, tomorrow)
+        results = api.content.find(Title=u'My event')
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].start, today)
+        self.assertEqual(results[0].end, tomorrow)
+
+    @unittest.skipUnless(HAS_PACONTENTYPES, 'Dexterity only')
+    def test_create_dx_event(self):
+        """https://github.com/plone/plone.app.contenttypes/issues/465"""
```
idgserpro: @hvelarde what's the relationship between plone/plone.app.contenttypes#465 and ```plone.api```? It seems to me more a problem of ```plone.app.contenttypes```.

I think this test can continue but without reference to ```plone.app.contenttypes```.

src/plone/api/tests/test_content.py
```diff
@@ -376,6 +376,64 @@ def test_create_at_with_title_in_request(self):
 
         self.assertEqual(page.title, 'Test document')
 
+    @unittest.skipIf(HAS_PACONTENTYPES, 'Archetypes only')
```
idgserpro: @gbastien thinking minor, I think this test should run with dx too. Collection creation should work with dx. You can remove this condition and change the name of the test.

src/plone/api/tests/test_content.py
hvelarde: IIRC, I was unable to reproduce it on a clean Plone installation; so maybe it's not related at all.

CHANGES.rst
```diff
@@ -21,6 +21,9 @@ Bug fixes:
 - fix typos in doc strings
   [tkimnguyen]
 
+- Fix falining AT Collection creation when using api.content.create.
```
hvelarde: typo

src/plone/api/tests/test_content.py
```diff
@@ -376,6 +376,64 @@ def test_create_at_with_title_in_request(self):
...
+    def test_create_at_collection(self):
+        """Test create at Collecition."""
```
hvelarde: typo

src/plone/api/tests/test_content.py
gbastien: I fixed the typos and run test for AT and DX and Travis is green.

I kept the test regarding the event creation, do not know if it is necessary now that we have a test with Collection creation for AT and DX?

idgserpro: @gbastien this event testing is necessary to avoid regression #364. I think it is necessary now only to remove the reference:

https://github.com/plone/plone.app.contenttypes/issues/465

idgserpro: In fact what avoids #364 is the AT test. But it's good to have an event creation test with DX

gbastien: OK so I would remove the test_create_at_event, unnecessary now because of test_create_collection?

idgserpro: @gbastien no. test_create_at_event avoid regression #364. test_create_collection not avoid that.

gbastien: Right, so let's merge? :-)

idgserpro: @gbastien I think it is necessary now only to remove the reference:

plone/plone.app.contenttypes#465

in docstring.

hvelarde:
@gbastien you need to run the Jenkins jobs also.

src/plone/api/tests/test_content.py
```diff
@@ -415,7 +415,7 @@ def test_create_at_event(self):
 
     @unittest.skipUnless(HAS_PACONTENTYPES, 'Dexterity only')
     def test_create_dx_event(self):
-        """https://github.com/plone/plone.app.contenttypes/issues/465"""
+        """ """
```
hvelarde: it desn't make sense to have an empty doctring.

src/plone/api/tests/test_content.py
```diff
@@ -376,6 +376,63 @@ def test_create_at_with_title_in_request(self):
...
+        self.assertEqual(event.startDate, today)
+        self.assertEqual(event.endDate, tomorrow)
+        results = api.content.find(Title=u'My event')
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].start, today)
+        self.assertEqual(results[0].end, tomorrow)
+
+    @unittest.skipUnless(HAS_PACONTENTYPES, 'Dexterity only')
+    def test_create_dx_event(self):
+        """ """
```
hvelarde: it doesn't make sense to have an empty docstring.

gbastien:
@hvelarde ready to merge ? ;-)

hvelarde:
I prefer to leave it to one @plone/framework-team member.

gforcada:
@hvelarde done: https://pypi.org/project/plone.api/ 1.8.6

hvelarde:
@gforcada in fact, it's 1.8.5.

PR #433, opened 2019-08-31
Mr tango/context state

MrTango:
Add new context module to give easy access to often used helpers from plone_context_stae view.


stevepiercy (changes requested):
We now use MyST instead of reST, and doc files should end in .md.

Additionally we no longer use the GitHub-only admonition.

PR #436, opened 2020-02-13
In tests, use stronger password.

mauritsvanrees:
Otherwise the tests in test_env.py fail during setup when the January 2020 hotfix is merged to CMFPlone.
The password strength is tested in more places then.


mauritsvanrees:
See for example [here on Jenkins](https://jenkins.plone.org/job/pull-request-5.1/218/testReport/junit/plone.api.tests.test_env/TestPloneApiEnv/test_adopt_fewers_role/)

mauritsvanrees:
- To me there is no sense in running all those Jenkins tests for this simple change
- Travis mostly succeeds after a minor update to `travis.yml`. 
- But the Travis test for the docs fails, which is totally unrelated to this PR.

So my inclination is to merge this. But I would like someone else to agree.

mauritsvanrees:
Issue for the failing Travis docs: https://github.com/plone/plone.api/issues/437

3021.bugfix
```diff
@@ -0,0 +1 @@
+In tests, use stronger password.  [maurits]
```
gforcada: Is that a path typo, or done in purpose? :thinking: this file should go to the `news` folder, isn't it?

mauritsvanrees: Oops. :-) Fixed.

ale-rt:

PR #449, opened 2021-02-01
Add module to handle relations

pbauer:
Integrate code from https://github.com/collective/collective.relationhelpers into plone.api
Fixes https://github.com/plone/Products.CMFPlone/issues/3137

### DIFF
--- a/docs/relation.rst
+++ b/docs/relation.rst
@@ -13,8 +13,6 @@
 Relations
 =========
 
-Note: the relations code only works for Dexterity content, not for Archetypes.
-
 To use the relations code, you should include `plone.api[relations]` in the dependencies of your project.
 
 .. _relation_create_example:
--- a/docs/relation.rst
+++ b/docs/relation.rst
@@ -36,6 +36,7 @@ Delete relation
 
     api.relation.delete(source=source, target=target, relationship="friend")
 
+In order to delete relation(s), you must provide either the ``source``, ``target`` or ``relationship``.
 
 Get relations
 =============
--- a/docs/relation.rst
+++ b/docs/relation.rst
@@ -13,6 +13,8 @@
 Relations
 =========
 
+Note: the relations code only works for Dexterity content, not for Archetypes.
+
 .. _relation_create_example:
 
 Create relation
--- a/docs/relation.rst
+++ b/docs/relation.rst
@@ -45,14 +45,15 @@ Get relations
 
     api.relation.get(source=source, target=target, relationship="friend", unrestricted=False, as_dict=False)
 
-You must provide either source, target or relationship, ``unrestricted`` and ``as_dict`` are optional.
+You must provide either source, target or relationship, or a combination of those.
+``unrestricted`` and ``as_dict`` are optional.
 
-Use the ``as_dict`` parameter if you want the result to be returned as a dictionary.
+By default the result is a list of objects.
+If you set ``as_dict=True`` it will return a dictionary with the names of the relations as keys and lists of objects as values.
 
-Use the ``unrestricted`` parameter if you want to bypass the View permission check on the relation objects(to/from_object).
-
-By default it returns a list.
-If source, target and relationship are all given, returns a list with a single item.
+By default the View permission is checked on the relation objects.
+You only get objects that you are allowed to see.
+Use the ``unrestricted`` parameter if you want to bypass this check.
 
 To get back relations, so relations pointing to an item, use:
 
--- a/docs/relation.rst
+++ b/docs/relation.rst
@@ -13,7 +13,42 @@
 Relations
 =========
 
-To use the relations code, you should include `plone.api[relations]` in the dependencies of your project.
+
+.. _relation_get_example:
+
+Get relations
+=============
+
+.. code-block:: python
+
+    api.relation.get(source=source, target=target, relationship="friend", unrestricted=False, as_dict=False)
+
+You must provide either source, target or relationship, or a combination of those.
+``unrestricted`` and ``as_dict`` are optional.
+
+By default the result is a list of ``RelationValue`` objects.
+
+If you set ``as_dict=True`` it will return a dictionary with the names of the relations as keys and lists of objects as values.
+
+By default the View permission is checked on the relation objects.
+You only get objects that you are allowed to see.
+Use the ``unrestricted`` parameter if you want to bypass this check.
+
+To get back relations, so relations pointing to an item, use:
+
+.. code-block:: python
+
+    api.relation.get(target=target)
+
+To get the objects connected by relations you can use the api of these return values:
+
+.. code-block:: python
+
+    for relation in api.relation.get(source=source):
+        source = relation.from_object
+        target = relation.to_object
+        relationship = relation.from_attribute
+
 
 .. _relation_create_example:
 
@@ -31,22 +66,21 @@ To create a relation between source object and target object, use :meth:`api.rel
     target = portal.bobby
     api.relation.create(source=source, target=target, relationship="friend")
 
+If the relation is based on a ``RelationChoice`` or ``RelationList`` field on the source object, the value of that field is created/updated accordingly.
+
+
+.. _relation_delete_example:
+
 Delete relation
 ===============
 
-Delete one specific relationship:
+Delete one or more relations:
 
 .. code-block:: python
 
     api.relation.delete(source=source, target=target, relationship="friend")
 
-You can delete all relations by explicitly asking:
-
-.. code-block:: python
-
-    api.relation.delete(delete_all=True)
-
-In all other cases, in order to delete relation(s), you must provide either ``source``, ``target`` or ``relationship``.
+In order to delete relation(s), you must provide either ``source``, ``target`` or ``relationship``.
 You can mix and match.
 
 Delete all relations from source to any target:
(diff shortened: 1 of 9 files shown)


docs/relation.rst
```diff
@@ -0,0 +1,67 @@
...
+    api.relation.get(source=source, target=target, relationship="friend", unrestricted=False, as_dict=False)
+
+You must provide either source, target or relationship, ``unrestricted`` and ``as_dict`` are optional.
+
+Use the ``as_dict`` parameter if you want the result to be returned as a dictionary.
+
+Use the ``unrestricted`` parameter if you want to bypass the View permission check on the relation objects(to/from_object).
+
+By default it returns a list.
+If source, target and relationship are all given, returns a list with a single item.
```
jensens, line 55: This is confusing, an example of both would help.

src/plone/api/relation.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # Plone 4 corrected paths
+    from plone.app.linkintegrity.handlers import modifiedDexterity as modifiedContent
+    from plone.app.linkintegrity.handlers import referencedRelationship
+
+
+logger = logging.getLogger(__name__)
+
+
+def _get_field_and_schema_for_fieldname(field_id, fti):
```
jensens, line 48: DRY: this function is already in the controlpanel PR https://github.com/plone/Products.CMFPlone/pull/3232 - see also my comments there.

src/plone/api/relation.py
```diff
@@ -0,0 +1,230 @@
...
+        # Iterate relations use a subclass of RelationValue
+        relation = StagingRelationValue(to_id)
+        event._setRelation(source, ITERATE_RELATION_NAME, relation)
+        return
+
+    fti = queryUtility(IDexterityFTI, name=source.portal_type)
+    if not fti:
+        logger.info(u'{} is no dexterity content'.format(source.portal_type))
+        return
+    field_and_schema = _get_field_and_schema_for_fieldname(from_attribute, fti)
```
jensens, line 109: also DRY - controlpanel
also see my comments about FTI in controlpanel PR

src/plone/api/relation.py
```diff
@@ -0,0 +1,230 @@
...
+    elif isinstance(field, (Relation, RelationChoice)):
+        logger.info('Add relation {} from {} to {}'.format(
+            from_attribute, source.absolute_url(), target.absolute_url()))
+        setattr(source, from_attribute, RelationValue(to_id))
+        modified(source)
+        return
+
+    # We should never end up here!
+    logger.info('Warning: Unexpected relation {} from {} to {}'.format(
+        from_attribute, source.absolute_url(), target.absolute_url()))
```
jensens, line 138: if we never should end up here, isn't this at least a warning or error situation to raise an api exception?

src/plone/api/relation.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+# @at_least_one_of('source', 'target', 'relationship')
```
jensens, line 141: ?

src/plone/api/relation.py
```diff
@@ -0,0 +1,230 @@
...
+                else:
+                    results.append(relation)
+            else:
+                continue
+        else:
+            if as_dict:
+                results[relation.from_attribute].append(relation)
+            else:
+                results.append(relation)
+    return results
```
jensens, line 230: DRY: I think I saw something very similar in the controlpanel code.

jensens (changes requested):
I love the get this features in API. A missing piece, really!

This one need coordination with https://github.com/plone/Products.CMFPlone/pull/3232. 
I would propose to get the other PR done, including factoring out utility low level stuff in an own module and avoid repeating complex code doing the same in two places.

mauritsvanrees:
@jensens said:

> This one need coordination with [plone/Products.CMFPlone#3232](https://github.com/plone/Products.CMFPlone/pull/3232).
> I would propose to get the other PR done, including factoring out utility low level stuff in an own module and avoid repeating complex code doing the same in two places.

So you say: coordinate this with the control panel, use code from there. More specifically: use its `relationhelper.py` file. Then we get into the question: for which Plone version is this?
For `plone.api` it sounds logical to me to include it in 5.2, so still supporting Python 2.7. But I see the `relationhelper.py` file already contains Python 3 code, at least f-strings.

I see @Petchesi-Iulian added some `except ImportError` code to still even support Plone 4.3, probably because tox and Travis still try to run it. That is not needed, I would say. Plone 4.3-5.1 is not supported anymore.

I would prefer to say: make this feature for Plone 5.2 and 6.0. `collective.relationhelpers` supports 5.2 with Python 2.7, so if both this PR and the control panel PR copy code from there, it should work. The control panel code is probably isolated enough that as release manager I could be persuaded to include it in 5.2, though I am not sure. Or maybe only the `relationhelpers.py` file (plus tests). But as said, that will already give problems because this file is Python 3 only...

Or do we say: this is only for Plone 6 and on 5.2 you can use `collective.relationhelpers`?

docs/relation.rst
mauritsvanrees: I think the original idea was to return a single item when source, target and relationship are all given. This is not the case in the current code: you always get a list (unless you use `as_dict`).
I clarified the docs.

src/plone/api/relation.py
mauritsvanrees: Comments there were hidden by now, but I found it:

> it is OK to use a portal_type string as FTI here, DX schema cache handles lookup transparent in high-performance way. So all calls of this function can save a getUtility call, which may speed up the whole process a bit and saves CPU cycles, time and CO2.

Done.

src/plone/api/relation.py
mauritsvanrees: Originally we had this restriction of needing at least one of source/target/relationship when deleting, but later I thought: when nothing is specified, we can delete all relations.
Might be a bit too implicit though. We can introduce a parameter `delete_all`. (`all` is a reserved keyword.) That seems nicer. I will do that.

src/plone/api/relation.py
mauritsvanrees: According to Philip the similar code in the controlpanel is from an earlier version, and the code is no longer there.

src/plone/api/relation.py
mauritsvanrees: The control panel is for Plone 6, and we want to have `plone.api.relation` in Plone 5. So we cannot share the code.

src/plone/api/relation.py
mauritsvanrees: You end up here when you create a relationship with a name that matches a field name, but is not a RelationList/Relation/RelationChoice. That could happen. Say you have an add-on `example.makefriends`. This does nothing with fields or schemas, it only wants to create relationships, so it calls `create(source=source, target=target, relationship="friend")`. Unknown to this add-on, source has a field `friend` which is a simple StringField. The best thing to do, is probably that the relation is created.
It *is* a bit unexpected though, and it *could* actually be an error, so a warning would be good as well.
I will do both then.

mauritsvanrees:
I have merged master in here, so tox and Travis are happier. `tox -e lint` rightly complains about a TODO item, and this is indeed not finished yet.

pbauer:
The relation-controlpanel (https://github.com/plone/Products.CMFPlone/pull/3232) is now merged into Plone 6. The `relationhelper.py` in CMFPlone has no code that is used or needed for this PR in plone.api.

I would prefer to make a version 2.0 of plone.api for Plone 6+ and py3 only. That would also allow us to remove all py2 and AT-compatability code from plone.api. And advise developers to use `collective.relationhelpers` in older versions.

mauritsvanrees:
> I would prefer to make a version 2.0 of plone.api for Plone 6+ and py3 only. That would also allow us to remove all py2 and AT-compatability code from plone.api. And advise developers to use `collective.relationhelpers` in older versions.

That makes sense to me.

Sorry, I have not worked on this since March. It is very useful, but it is not very high on my priority list.

pbauer:
Regarding `relations.delete()`: 
* I think we should remove RelationValues form the source objects if they exist. 
* We should not add the `delete_all` option. I don't think it falls in the 20/80-rule and it already exists in `Products.CMFPlone.relationhelper.purge_relations` where it makes more sense in connection with `rebuild_relations`.

pbauer:
I finished delete as I proposed above and added some more tests. This is now ready to be reviewed.

This branch is not on top of https://github.com/plone/plone.api/pull/461 in case we will go ahead with creating a 2.0 release without AT and py2 support. If we decide not to do that not I will re-add support for py2 and Plone 5.2 (but obviously not support for AT).

pbauer:

jensens (approved):

ale-rt:
I rebased this branch on master to have a clean diff and fix some minor details.

It looks good to me, even if there are some small details that should be fixed.
```
[ale@emily plone.api]$ flake8 src/|grep relation
src/plone/api/tests/test_relation.py:189:5: E303 too many blank lines (2)
src/plone/api/tests/test_relation.py:345:89: E501 line too long (90 > 88 characters)
src/plone/api/tests/test_relation.py:346:89: E501 line too long (89 > 88 characters)
src/plone/api/tests/test_relation.py:364:89: E501 line too long (93 > 88 characters)
src/plone/api/tests/test_relation.py:423:89: E501 line too long (90 > 88 characters)
```
Not a big deal, we maybe can handle this later by running black on the whole code base.

I would recommend to release the package initially as an alpha to allow the new methods signature to be changed without the need to bump the version number.

ale-rt (approved):

jensens:

pbauer:
Thanks for merging 🎉

jensens:
Thanks for the work!

Issue #450, opened 2021-02-02
Drop Plone Support < 5.2


jensens:
I propose to drop support of Plone < 5.2 for plone.api. Otherwise adding of new features gets very complex (looking at the relations PR).

mauritsvanrees:
@plone/framework-team Is anyone against this?

I have gone ahead and made a branch 1.10.x. For good measure I have updated coredev 4.3 and 5.1 to use this.

ale-rt:
Some consideration:
- I think we can easily drop AT support and keep support for Plone 5.2.
- I am not against dropping support for Python 2 (I bet it will work well enough on Plone 5.2 + Python3 as well, at least as long as the planned relations module is not used).
- if needed I think some fixes can be backported to 1.10.x quite easily

Personally I would say go ahead.
But I would really like to hear other opinions from @plone/framework-team

ale-rt:
Sorry I misread the title :)
That said my "go ahead" is even more firm!

jensens:
Dropping Plone 5.2 in combination of dropped Py 2.7 and dropped AT support in a 2.x series sounds like a plan, while keeping the 1.x series supporting 4.3, 5.1 and 5.2 on Py2.7 and with AT.

mauritsvanrees:
> Dropping Plone 5.2 in combination of dropped Py 2.7 and dropped AT support in a 2.x series sounds like a plan.

Sounds confusing. This looks like dropping 5.2 support is different from dropping Py 2 and AT. But maybe I just misinterpret you.
I don't want to drop 5.2 support yet. The plan is to add the relations support to 5.2, which means supporting Python 2.7. Should be easy enough, because as far as I know the collective.relationhelpers code works on Py 2.7.
So my preference:

- 1.10.x: support 4.3-5.2, Py 2+3
- 1.11.x: add relation module, support 5.2-6.0, Py 2+3
- master/2.x: only support 6.0 Py 3.

mauritsvanrees:
As expected, it seems there is no opposition to dropping support for Plone 5.1 and earlier. I have merged my changes to that effect in [PR 431](https://github.com/plone/plone.api/pull/431#pullrequestreview-603947042). Once it is green, for me it is fine to merge.

PR #431, opened 2019-06-17
Tox improvements and added black as formatter

loechel:

PR #453, opened 2021-02-28
Further improvements to the tox PR

mauritsvanrees:
This is a PR for merging into the branch of PR https://github.com/plone/plone.api/pull/431

@loechel I already merged master into your branch, because there were two minor conflicts.

For this PR, see the individual commits.
It also drops support for Plone 5.1 and earlier. See also https://github.com/plone/plone.api/issues/450

Locally, the 2.7 tests fail, but that has a weird cause. I will open an issue for that.
The other tox tests pass.


thet:
@mauritsvanrees can this be merged to master as soon as the travis failures are resolved?

and should we also switch to github actions? I'm thinking of adding support for that.

mauritsvanrees:
I would like to use GitHub Actions, yes. The plug on travis-co.org has not been pulled yet, but that is a matter of time.

This part of the PR drops testing on 5.1 and earlier, which should be okay, but we may want to wait on a resolution of the discussion in https://github.com/plone/plone.api/issues/450

Other than the supported versions, biggest question of this PR and the parent PR of Alexander, is if the `docs` tox env still works, and if this is needed for building the documentation on docs.plone.org. Currently it fails on Travis, so maybe there is some more work to do.

mauritsvanrees:
Travis is green. I merge to the PR branch of @loechel

PR #455, opened 2021-03-04
Tox: only test docs, verbose, and run with 3.7, 3.8 and 3.9.

mauritsvanrees:
On Travis the sphinx docs fail with no output after ten minutes.

This PR not meant to be merged, this is only for investigation.
Based on https://github.com/plone/plone.api/pull/453

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,6 @@ matrix:
   include:
       - python: "3.7"
         env: TOXENV=docs
-      - python: "3.8"
-        env: TOXENV=docs
-      - python: "3.9"
-        env: TOXENV=docs
 
 before_install:
     - mkdir -p $HOME/buildout-cache/{downloads,eggs,extends}
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,20 +10,12 @@ global:
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Python 3.7 works, but with 3.8 and 3.9 it still hangs at simply installing `plone.api` with pip:

```
$ travis_retry tox
docs create: /home/travis/build/plone/plone.api/.tox/docs
docs installdeps: Sphinx
docs develop-inst: /home/travis/build/plone/plone.api

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

To properly build the documentation including showing the sources, the plone.api code needs to be importable, which means we need to actually install all Plone code. I guess with pip this still not always works.

I see the same problem locally. I tried explicitly installing plone.api with pip and `constraints.txt` from Plone 5.2.4, and more verboseness, but this also hangs after a while, my CPU is maxed out, and I see the Python process is busy, but I don't see what it is doing.
This morning I think I saw it work locally with various Python versions, but maybe tox had enough files cached. Now it only works with Python 2.7. Yes, 2 not 3.

Meanwhile, Travis passes nicely when using 3.7, so let me close this PR and update the original.

PR #461, opened 2021-08-22
Drop Archetypes and Python 2

### LINKED ISSUE #460: Create 2.0 without support for Python 2 and Archetypes
As mentioned in https://github.com/plone/plone.api/pull/449 I propose to have a plone.api version 2.0 that only targets Plone 6 and above. The code and tests could be cleaned up and simplified in many places. 
Also nobody would expect the new api methods for relations to work with Archetypes. 

TODOs:
- [x] Cleanup code to remove Archetypes-specific code
- [x] Update code (e.g. using pyupgrade) to only support Python 3
- [ ] Document changes

pbauer:
Fixes #460


pbauer:

src/plone/api/tests/test_content.py
```diff
@@ -421,46 +410,6 @@ def test_create_collection(self):
...
-            endDate=tomorrow,
-        )
-        self.assertEqual(event.startDate, today)
-        self.assertEqual(event.endDate, tomorrow)
-        results = api.content.find(Title=u'My event')
-        self.assertEqual(len(results), 1)
-        self.assertEqual(results[0].start, today)
-        self.assertEqual(results[0].end, tomorrow)
-
-    @unittest.skipUnless(HAS_PACONTENTYPES, 'Dexterity only')
```
ale-rt, line 444: I think this test should stay

tox.ini
```diff
@@ -34,7 +31,6 @@ commands =
 
 setenv =
     COVERAGE_FILE=.coverage.{envname}
-    plone-5.2-archetypes: BUILDOUT_FILE="plone-5.2.x-archetypes.cfg"
     plone-5.2: BUILDOUT_FILE="plone-5.2.x.cfg"
```
ale-rt, line 34: The tox.ini needs some love, but tox tests are failing anyway and switching to GHA might be a good idea. 
So I would not bother about it.

.travis.yml
```diff
@@ -14,10 +14,6 @@ matrix:
         env: TOXENV=isort,lint
       - python: "3.7"
         env: TOXENV=docs
-      - python: "2.7"
-        env: TOXENV=py27-plone-5.2
-      - python: "2.7"
-        env: TOXENV=py27-plone-5.2-archetypes
       - python: "3.6"
         env: TOXENV=py36-plone-5.2
```
ale-rt, line 18: Probably it would be good to move this to GHA once we have a Plone 6 alpha released

ale-rt:
It would also good to remove mock from the requirements, but that can be done later/independently https://github.com/plone/plone.api/pull/461/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R46

src/plone/api/tests/test_content.py
pbauer: True, I brought it back.

pbauer:
`mock` can be removed from dependencies. Only `from unittest import mock` is used in tests now.

ale-rt (approved):
Thanks a lot!

ale-rt:
I added the @plone/framework-team members as reviewers, I would say if you got 3 positive reviews click the merge button.

It goes without saying that this is open to be reviewed by whoever wants!

pbauer:
@plone/framework-team can we please get a third opinion?

Issue #462, opened 2021-09-15
Relations: Plone fails to start up when plone.app.iterate is missing


mauritsvanrees:
The recently added relation module has an undeclared dependency on `plone.app.iterate`. When starting up Plone 6 with only `Products.CMFPlone`, I get an error:

```
  File "/Users/maurits/shared-eggs/cp39/plone.app.discussion-4.0.0a2-py3.9.egg/plone/app/discussion/browser/moderation.py", line 6, in <module>
    from plone import api
  File "/Users/maurits/shared-eggs/cp39/plone.api-2.0.0a1-py3.9.egg/plone/api/__init__.py", line 7, in <module>
    from plone.api import relation
  File "/Users/maurits/shared-eggs/cp39/plone.api-2.0.0a1-py3.9.egg/plone/api/relation.py", line 10, in <module>
    from plone.app.iterate.dexterity import ITERATE_RELATION_NAME
zope.configuration.xmlconfig.ZopeXMLConfigurationError: File "/Users/maurits/shared-eggs/cp39/plone.app.discussion-4.0.0a2-py3.9.egg/plone/app/discussion/browser/configure.zcml", line 16.4-22.10
    File "/Users/maurits/tmp/test/parts/instance/etc/site.zcml", line 16.2-16.23
    File "/Users/maurits/shared-eggs/cp39/Products.CMFPlone-6.0.0a1.dev0-py3.9.egg/Products/CMFPlone/configure.zcml", line 16.2-16.44
    File "/Users/maurits/shared-eggs/cp39/plone.app.discussion-4.0.0a2-py3.9.egg/plone/app/discussion/configure.zcml", line 33.4-33.34
    ModuleNotFoundError: No module named 'plone.app.iterate'
```

See also https://github.com/plone/plone.app.discussion/issues/188, because plone.app.discussion should not be using plone.api.

PR #463, opened 2021-09-27
Prevent startup error in relations when plone.app.iterate is missing

mauritsvanrees:
Fixes https://github.com/plone/plone.api/issues/462


mauritsvanrees:

PR #470, opened 2022-04-06
Conversion to myst markdown

ksuess:


mauritsvanrees:
@ksuess I only now realise that this has an unwanted side effect: lots of tests are gone.

I started writing more here, but a new issue is better. See issue #474.

PR #484, opened 2022-05-18
Use the dexterity machinery in api.content.create

### LINKED ISSUE #439: In api.content.create, avoid renaming object if "id" was not given
Hi @jensens @ale-rt 

we discovered that when creating an element using content.create without specifying an "id", it was created then renamed, this cause ContainerModified event to be called 2 times (and some other events too).

I wondered why this was done in 2 steps?  Could it be done before creating the element?
https://github.com/plone/plone.api/blob/master/src/plone/api/content.py#L118
Actually the element is created and if id is not given, then it is renamed to a computed id.
I think we could check if "id" is given, then generate it and pass it to invokeFactory instead renaming it after?
Or is there some reason why it is done after?

We could propose a PR "solving" this.

We avoid this by passing a generated id to content.create now.

Thank you,

Gauthier

ale-rt:
Fixes #439


ale-rt:

mauritsvanrees (commented):
So far, it looks good to me, from looking at the code.
Nice one with the context manager in the tests. :-)

ale-rt:
> So far, it looks good to me, from looking at the code. Nice one with the context manager in the tests. :-)

Thanks for looking at it.
I still see some issues, e.g. this test passes in master but not in this branch:
```
+     def test_create_document_subject(self):
+         "When passing the subject as a string, it should be saved as a tuple"
+         obj = api.content.create(
+             container=self.portal,
+             type="Dexterity Item",
+             id="foo",
+             subject="bar"
+         )
+         self.assertEqual(obj.subject, ("bar",))
```

src/plone/api/content.py
```diff
@@ -62,14 +61,50 @@ def create(
...
+                            "The id you provided conflicts with "
+                            "an existing object or it is reserved"
+                        )
+            else:
+                content.id = name_chooser.chooseName(title, content)
+
+            content_id = container._setObject(content.id, content)
+        else:
+            content_id = not safe_id and id or str(random.randint(0, 99999999))
+            container.invokeFactory(type, content_id, **kwargs)
```
davisagli, line 107: We should keep the renaming below in this case (for non-DX content), no?

ale-rt:
Sorry to have dropped the ball on this, but IIRC, I gave up because I found this patch breaking the ploneintranet build.
I do not remember the details but I can recheck if this is still the case.

PR #488, opened 2022-05-30
plone.api.content.get should always return a content.

### LINKED ISSUE #487: `plone.api.content.get(path)` returns a field if it finds one
Given a path like:

`path = '/de/examples/blocks`

and no content named blocks, it will get and return the `blocks` field content, because of `restrictedTraverse`...

Since `plone.api.content.get` promises to return a content type, we should not return anything that it's not.

ericof:
Fixes #487

### DIFF
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -9,6 +9,7 @@ from plone.api.validation import required_parameters
 from plone.app.linkintegrity.exceptions import LinkIntegrityNotificationException
 from plone.app.uuid.utils import uuidToObject
 from plone.uuid.interfaces import IUUID
+from Products.CMFCore.DynamicType import DynamicType
 from Products.CMFCore.WorkflowCore import WorkflowException
 from zope.component import getMultiAdapter
 from zope.component import getSiteManager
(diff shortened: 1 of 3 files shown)


ericof:

src/plone/api/content.py
```diff
@@ -130,7 +131,10 @@ def get(path=None, UID=None):
             )
 
         try:
-            return site.restrictedTraverse(path)
+            content = site.restrictedTraverse(path)
+            # Only return a content if it implements DynamicType,
+            # which is true for Dexterity content and Comment (plone.app.discussion)
+            return content if isinstance(content, DynamicType) else None
```
ale-rt, line 137: Generally it is better to keep the `try/except` as small as possible.

So something like this:
```python
try:
    content = site.restrictedTraverse(path)
except (KeyError, AttributeError):
    return None
if $CONDITION:
    return content
```

About `$CONDITION` I am pretty surprised by your choice of checking if content is an instance of the `DynamicType` mixin class.

Is there a reason for that?

Maybe given that we just don't want to return by mistake some object attribute, what about having the condition be:
```python:
from persistent import Persistent
if isinstance(content, Persistent):
    return content
```

ale-rt (approved):
LGTM

This is fine: it solves the original issue and I do not see any harm for the current adopters of `plone.api` (unless they are actually exploiting the fact that `content.get` can also traverse the objects attributes).

I left some comments that you can totally ignore.

ericof:

ericof:
@ale-rt, @mauritsvanrees Any idea why Robot framework tests keep failing for this PR?

stevepiercy:
@ericof Jenkins kept failing for me and @ksuess while working on docs last week due to this issue:
https://github.com/plone/Products.CMFPlone/issues/3542

I assume it is the same problem we had.

PR #493, opened 2022-09-03
Fix the security context managers

### LINKED ISSUE #374: plone.api.env.adopt_role does not remove security manager context when exception was raised within context manager
In this example you can see that we're making use of adopt_roles 100 times in a single script, the fallback max stack size as per https://github.com/zopefoundation/AccessControl/blob/master/src/AccessControl/ImplPython.py#L535

```    def test_blah(self):
        x = 0
        while x <= 101:
            x = x + 1
            try:
                with plone.api.env.adopt_roles(['Manager']):
                    raise ValueError('wat')
            except ValueError:
                pass
```

This raises the following error:
```Error in test test_blah
Traceback (most recent call last):
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "tests/test_blah.py", line 1023, in test_blah
    with plone.api.env.adopt_roles(['Manager']):
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "buildout-cache/eggs/plone.api-1.5.1-py2.7.egg/plone/api/env.py", line 118, in _adopt_roles
    sm.addContext(overriding_context)
  File "buildout-cache/eggs/AccessControl-3.0.11-py2.7-linux-x86_64.egg/AccessControl/ImplPython.py", line 604, in addContext
    raise SystemError, 'Excessive recursion'
SystemError: Excessive recursion
```

This can be fixed by amending this section of code (https://github.com/plone/plone.api/blob/master/src/plone/api/env.py#L120) to

```
    try:
        yield
    finally:
        sm.removeContext(overriding_context)
```

ale-rt:
Ensure that the security related context managers
restore the context even if an error occurs.

Fixes #374

### DIFF
--- a/news/374.bugfix
+++ b/news/374.bugfix
@@ -1,0 +1,2 @@
+Ensure that the security related context managers
+restore the context even if an error occurs.
--- a/src/plone/api/env.py
+++ b/src/plone/api/env.py
@@ -80,9 +80,10 @@ def _adopt_user(user):
     old_security_manager = getSecurityManager()
     newSecurityManager(getRequest(), user)
 
-    yield
-
-    setSecurityManager(old_security_manager)
+    try:
+        yield
+    finally:
+        setSecurityManager(old_security_manager)
 
 
 @required_parameters("roles")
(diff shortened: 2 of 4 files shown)


davisagli (approved):
Good fix, thanks.

PR #496, opened 2022-10-15
Trigger build and deploy of Plone 6 documentation

esteele:
Calls an Action on the Plone 6 documentation repo that updates the documentation registry's submodules and triggers a new build and deploy of the Plone 6 docs.

Refs plone/documentation#1214

### DIFF
--- a/.github/workflows/update_remote_docs.yml
+++ b/.github/workflows/update_remote_docs.yml
@@ -1,0 +1,19 @@
+name: Trigger build and deploy of Plone 6 documentation
+
+on:
+  push:
+    branches:
+      - "master"
+    paths:
+      - "docs/*"
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/update_remote_docs.yml
```diff
@@ -0,0 +1,19 @@
...
+    paths:
+      - "docs/*"
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          curl -X POST \
+          -H "Authorization: Bearer ${{secrets.DOCUMENTATION_BUILD_APPLICATION_KEY}}" \
```
davisagli, line 16: @esteele For future reference, where is this app defined? I thought I might find it in https://github.com/organizations/plone/settings/apps, but I didn't.

esteele: It's on the Plone Foundation user account, not the organization.

PR #497, opened 2022-12-21
Switch to 6.docs.plone.org (was 6.dev-docs.plone.org)

ksuess:


ksuess:
One task of https://github.com/plone/documentation/issues/1140

news/497.feature
```diff
@@ -0,0 +1,2 @@
+Switch to 6.docs.plone.org (was 6.dev-docs.plone.org)
```
stevepiercy, line 1: I would argue that this is not a new feature, but that's the best match out of what is available in the towncrier configuration.

I wish we had a better name for documentation changes.

https://github.com/plone/plone.api/blob/master/pyproject.toml#L12-L16

Should I make a new issue to request that addition for a future PR, not this one?

stevepiercy (approved):

news/497.feature
ksuess: Thought the same.

Would be nice to have a towncrier news entry type '.doc'.
I think plone.releaser is the place to suggest an addition. See https://github.com/plone/plone.releaser/blob/master/ADD-A-NEWS-ITEM.rst

ksuess:
Usually in Plone repos I do not merge my own PRs. Can you do that please, @mauritsvanrees 
You know what else is necessary for the big picture. Thank you!

news/497.feature
mauritsvanrees: Each project that uses `towncrier` has a `pyproject.toml` where the known suffixes are defined. For packages like `plone.api` from which we include the documentation on docs.plone.org, it makes sense to allow `.doc`. Maybe for all packages.
I will merge this PR, then copy the `pyproject.toml` from `plone.restapi` and fix the news snippet extension.

stevepiercy: I would prefer `.documentation` because PyCharm and macOS attempt to open `.doc` in Word or other Office app.

mauritsvanrees: done

Issue #501, opened 2023-03-28
Delete relations by relation name is failing


ksuess:
RuntimeError: the bucket being iterated changed size

Also deletion by source and target are failing, if non-field relation in results.

PR #502, opened 2023-03-28
Fix api.relation.delete by relation name, by source, by target (inste…

ksuess:
…ad of single relation defined by source, target, relation name)

Fix https://github.com/plone/plone.api/issues/501

### DIFF
--- a/src/plone/api/relation.py
+++ b/src/plone/api/relation.py
@@ -281,7 +281,8 @@ def delete(source=None, target=None, relationship=None):
         query["to_id"] = intids.getId(target)
     if relationship is not None:
         query["from_attribute"] = relationship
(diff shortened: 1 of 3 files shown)


ksuess:

src/plone/api/relation.py
```diff
@@ -292,7 +293,7 @@ def delete(source=None, target=None, relationship=None):
             # The relationship is not the name of a dexterity field.
             # Only purge relation from relation-catalog.
             relation_catalog.unindex(rel)
-            return
+            continue
```
davisagli, line 296: Good catch

davisagli (approved):
Thanks for the fix and new tests! LGTM.

PR #507, opened 2023-05-18
Create relation only if there is no existing one with same …

ksuess:
source, target, relationname.
But mark source as modified.

### DIFF
--- a/news/507.bugfix
+++ b/news/507.bugfix
@@ -1,0 +1,2 @@
+Create relation only if there is no existing one with same source, target, relationname.
+But mark source as modified. @ksuess
--- a/src/plone/api/relation.py
+++ b/src/plone/api/relation.py
@@ -170,9 +170,8 @@ def create(source=None, target=None, relationship=None):
         "from_id": from_id,
         "to_id": to_id,
     }
-    has_relation = False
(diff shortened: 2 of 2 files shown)


stevepiercy (commented):
It would be good to add a sentence or two to the documentation with this detail. I've also modified an existing sentence. I'm not sure if the new suggestion is 100% correct

https://6.docs.plone.org/plone.api/relation.html#create-relation

```
If the relation is based on a `RelationChoice` or `RelationList` field on the source object, then the value of that field is created or updated accordingly.

When calling {meth}`create()`, if a relation with the same `source`, `target`, or `relationship` does not exist, then it creates the relation and marks the `source` as modified. However if the relation exists, then [something else happens or doesn't].
```

ksuess:

pbauer (approved):

ksuess:
@mauritsvanrees  May this be released, please. It's needed for administering relations in the upcoming relations control panel https://github.com/plone/volto/pull/3382

mauritsvanrees:
Sure.  I have released `plone.api` 2.0.3.

ksuess:
> Sure. I have released `plone.api` 2.0.3.

Thank you!

PR #509, opened 2023-05-19
Config with default template

gforcada:
It's using https://github.com/plone/meta/pull/92


.meta.toml
```diff
@@ -0,0 +1,176 @@
...
+    *.rst
+    */doctests/*
+    */tests/*
+
+[coverage:html]
+directory = _build/coverage"""
+
+[tox]
+envlist_lines = """
+    py{38,39}-plone{52}
```
mauritsvanrees, line 51: We should not test Plone 5.2 on Python 3.9.  But that is already happening on current master I think, so could be handled separately.

.meta.toml
```diff
@@ -0,0 +1,176 @@
...
+
+setenv =
+    BUILDOUT_FILE=test_plone-60.cfg
+    plone52: BUILDOUT_FILE=test_plone-52.cfg
+
+deps =
+    pdbpp
+    manuel
+
+# tox < 4.0.0
```
mauritsvanrees, line 87: These lines can be removed, we require tox 4.4.0+ already.

mauritsvanrees (approved):
LGTM.  Two minor inline comments, but could be handled later as well.

gforcada:
Duplicated of #512

PR #510, opened 2023-05-24
Enhance API docs of portal.translate to show that the domain is optio…

thet:
…nal in some cases.

If anyone can point me to some documentation on the api doc syntax, I'd be happy. Didn't found anything than docs on how to use sphinx autodoc on an API level and so forth.

### DIFF
--- a/news/510.documentation
+++ b/news/510.documentation
@@ -1,0 +1,2 @@
+Enhance API docs of portal.translate to show that the domain is optional in some cases.
+[thet]
--- a/news/510.documentation
+++ b/news/510.documentation
@@ -1,2 +1,1 @@
 Enhance API docs of `portal.translate` to show that the domain is optional in some cases. @thet
-[thet]
--- a/news/510.documentation
+++ b/news/510.documentation
@@ -1,2 +1,2 @@
-Enhance API docs of portal.translate to show that the domain is optional in some cases.
+Enhance API docs of `portal.translate` to show that the domain is optional in some cases. @thet
 [thet]
--- a/src/plone/api/portal.py
+++ b/src/plone/api/portal.py
@@ -408,7 +408,7 @@ def translate(msgid, domain="plone", lang=None):
 
     :param msgid: [required] message to translate
     :type msgid: string | zope.i18nmessageid.Message
-    :param domain: i18n domain to use. When `msgid` is a instance of `Message`
+    :param domain: i18n domain to use. When ``msgid`` is an instance of ``Message``,
                    the Message' domain is used.
     :type domain: string
     :param lang: target language
--- a/src/plone/api/portal.py
+++ b/src/plone/api/portal.py
@@ -407,7 +407,8 @@ def translate(msgid, domain="plone", lang=None):
     Default to current negotiated language if no target language specified.
 
     :param msgid: [required] message to translate
-    :type msgid: string | zope.i18nmessageid.Message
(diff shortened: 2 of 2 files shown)


src/plone/api/portal.py
```diff
@@ -407,8 +407,9 @@ def translate(msgid, domain="plone", lang=None):
     Default to current negotiated language if no target language specified.
 
     :param msgid: [required] message to translate
-    :type msgid: string
-    :param domain: i18n domain to use
+    :type msgid: string | zope.i18nmessageid.Message
+    :param domain: i18n domain to use. When `msgid` is a instance of `Message`
+                   the Message' domain is used.
```
thet, line 412: Is the intend on this line right or will it fail any docstring formatters?

src/plone/api/portal.py
```diff
@@ -407,8 +407,9 @@ def translate(msgid, domain="plone", lang=None):
     Default to current negotiated language if no target language specified.
 
     :param msgid: [required] message to translate
-    :type msgid: string
-    :param domain: i18n domain to use
+    :type msgid: string | zope.i18nmessageid.Message
+    :param domain: i18n domain to use. When `msgid` is a instance of `Message`
```
stevepiercy, line 411: docstrings need to follow reST syntax, not MyST syntax. I know, annoying, isn't it?

```suggestion
    :param domain: i18n domain to use. When ``msgid`` is an instance of ``Message``,
```

src/plone/api/portal.py
stevepiercy, line 412:
```suggestion
                   then the ``Message``'s domain is used.
```

src/plone/api/portal.py
```diff
@@ -407,8 +407,9 @@ def translate(msgid, domain="plone", lang=None):
     Default to current negotiated language if no target language specified.
 
     :param msgid: [required] message to translate
-    :type msgid: string
-    :param domain: i18n domain to use
+    :type msgid: string | zope.i18nmessageid.Message
```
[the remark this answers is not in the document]
stevepiercy: I think it is the following, based on this example: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-signatures

```suggestion
    :type msgid: string
    :type msgid: zope.i18nmessageid.Message
```

stevepiercy (changes requested):
I don't know for sure, but let's try these suggestions. We'll figure it out.

thet:
@stevepiercy tnx for the review and sorry for my late reply.
I've updated this PR with your suggestions.

news/510.documentation
```diff
@@ -0,0 +1,2 @@
+Enhance API docs of portal.translate to show that the domain is optional in some cases.
```
stevepiercy, line 1: Sprinkle some MyST in there, standardize author.

```suggestion
Enhance API docs of `portal.translate` to show that the domain is optional in some cases. @thet
```

news/510.documentation
```diff
@@ -0,0 +1,2 @@
...
+[thet]
```
stevepiercy, line 2:
```suggestion
```

stevepiercy (changes requested):
Looks good, just a little tweak for the change log, and it is good to go. Thank you!

thet:
Thank you! I applied your suggestions. I'm merging.

PR #512, opened 2023-06-25
Config with default template

gforcada:


gforcada:
Finally this is working 🎉 the coverage GHA was failing due to a setting in `setup.cfg` 😕 

We should probably clean up the configuration files, as there are quite a mix of things: buildout/tox/make... and probably outdated configuration all over 😓  but let's do that on a separate PR

gforcada:

PR #513, opened 2023-07-11
Do not run GitHub Actions tests twice.

thet:
Only run GitHub Actions tests when commiting directly against master or main or opening a pull request agains master or main. This avoids to run the same test suite for the same environment twice.

### DIFF
--- a/.github/workflows/plone_python.yml
+++ b/.github/workflows/plone_python.yml
@@ -2,7 +2,13 @@ name: Plone backend tests
 
 on:
   push:
+    branches:
+      - master
+      - main
   pull_request:
+    branches:
+      - master
+      - main
 
 jobs:
   plone_python:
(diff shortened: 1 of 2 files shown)


davisagli (approved):

wesleybl:
I think it's bad because the test will only run in PR. I like to check the tests before doing the PR. I prefer the approach used in Volto:

https://github.com/plone/volto/blob/master/.github/workflows/acceptance.yml#L5

davisagli:
@wesleybl You could open a draft PR and mention that you are opening it in order to run the tests. But, I would also be okay with the way we did it for volto and plone.restapi.

thet:
@davisagli Tnx for the review

@wesleybl I'm merging but as david said, you can open another PR. 
I personally don't need test runs on branches without a PR. I open PRs early in "draft" state. That's more visible not only to others but also to me which reduces the chance to forget about that work.

If every project on GitHub would optimize the test runs we would save some significant amount of energy :green_heart:

PR #514, opened 2023-10-05
Fix regex asserts

gforcada:
That's part of https://github.com/plone/Products.CMFPlone/issues/3852 effort

### DIFF
--- a/news/1.bugfix
+++ b/news/1.bugfix
@@ -1,0 +1,2 @@
+Replace deprecated assert methods.
+[gforcada]
--- a/src/plone/api/tests/test_env.py
+++ b/src/plone/api/tests/test_env.py
@@ -514,14 +514,14 @@ class TestPloneApiEnv(unittest.TestCase):
         from plone.api.env import plone_version
 
         self.assertTrue(isinstance(plone_version(), str))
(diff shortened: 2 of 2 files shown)


jensens:
Why are there still tests for 5.2? Should we run `plone.meta` on this before merging here?

mauritsvanrees:
Coredev 5.2 uses branch 1.x, so from that point of view it is fine to drop 5.2 support.

gforcada:
@jensens @mauritsvanrees on #519 I fixed the plone/python versions being tested, so that PRs can actually pass 👍🏾 

This one should work now 🍀

gforcada:

gforcada:
@jensens `plone/meta` is on #509 I will rebase it and there were a few things that needed discussion, or that I was not 100% sure on how to handle them

gforcada:
The failing 3.12 job is expected, as we are missing some version pins from https://github.com/plone/buildout.coredev/pull/890 which in turn fails on 3.12 because it's missing the fixes here 🥚 🐔

gforcada:
Merging as it works on 6.0/1 and it is needed for Python 3.12 👍🏾

PR #518, opened 2023-10-13
Fix failing tests related to object_provides index

Akshat2Jain:
this pr is related to [#3836](https://github.com/plone/Products.CMFPlone/pull/3836)


src/plone/api/tests/test_content.py
```diff
@@ -1014,7 +1014,7 @@ def test_find_interface_dict(self):
                 "operator": "and",
             },
         )
-        self.assertEqual(len(brains), 1)
+        self.assertEqual(len(brains), 2)
```
wesleybl, line 1017:
```suggestion
        # Plone Site also implements the IContentish and INavigationRoot interfaces.
        self.assertEqual(len(brains), 2)
```

wesleybl: The assert on line 1026 also needs to be fixed:

https://github.com/plone/plone.api/blob/6504c3f277e09917627179fbacb102809679f467/src/plone/api/tests/test_content.py#L1026

src/plone/api/tests/test_content.py
```diff
@@ -1042,7 +1042,7 @@ def test_find_interface_dict__include_not_query(self):
             },
         )
 
-        self.assertEqual(len(brains_all) - len(brains), 1)
+        self.assertEqual(len(brains_all) - len(brains), 0)
```
wesleybl, line 1045: The assert here should be 2. You can see the error in the test here:

https://jenkins.plone.org/job/pull-request-6.1-3.11/298/testReport/junit/plone.api.tests.test_content/TestPloneApiContent/test_find_interface_dict__include_not_query/

```suggestion
         # Plone Site also implements the IContentish and INavigationRoot interfaces.
        self.assertEqual(len(brains_all) - len(brains), 2)
```

wesleybl (changes requested):
Please add a line in the PR description saying it needs: https://github.com/plone/Products.CMFPlone/pull/3856

wesleybl:
@Akshat2Jain Please click on the link `Plone Jenkins CI - pull-request-6.1-3.11` below and put the link this PR and the PR https://github.com/plone/Products.CMFPlone/pull/3836, so that you can test them be executed with these two PR.

And in the Plone 6.0 tests, put the link to this PR and https://github.com/plone/Products.CMFPlone/pull/3856

src/plone/api/tests/test_content.py
```diff
@@ -1041,8 +1042,8 @@ def test_find_interface_dict__include_not_query(self):
                 "not": INavigationRoot.__identifier__,
             },
         )
-
-        self.assertEqual(len(brains_all) - len(brains), 1)
+        # Plone Site also implements the IContentish and INavigationRoot interfaces.
+        self.assertEqual(len(brains_all) - len(brains), 2)
```
davisagli, line 1046: @wesleybl This couples plone.api closely to the version of CMFPlone (i.e. a new version of plone.api will require a new version of CMFPlone, and an old version of plone.api will require an old version of CMFPlone...at least in order to have these tests pass.

To make it less tightly coupled, I think you could add `"not": IPloneSiteRoot.__identifier__` to the query, and continue to check for 1 result. Then the test will work with and without the update to the Plone site interfaces.

Akshat2Jain: It seems we are initially getting all objects that provide the IContentish interface and then applying a 'not' query to exclude those that provide the INavigationRoot interface. The assertion should be updated to check that there is one less object in the result compared to the initial count. The updated assertion should look like self.assertEqual(len(brains_all) - len(brains), 1). is that right @davisagli @wesleybl?

davisagli: @Akshat2Jain Oh yeah, comparing the final count to the initial count would also work. Good idea

wesleybl: @davisagli good idea! The test `test_find_interface_dict` would have to be fixed too.

> The updated assertion should look like self.assertEqual(len(brains_all) - len(brains), 1)

@Akshat2Jain Plone Site also implements the `INavigationRoot` interface. So a search with `"not": INavigationRoot.__identifier__` will also exclude the Plone Site. So we will have two less.

I think the best option is @davisagli 's suggestion. The only "problem" with this option is that if new content is created in the test class, the test will fail and will have to be fixed. But I don't think that's such a big problem.

Akshat2Jain: Should I undo the change then?

wesleybl: @Akshat2Jain yes. You should leave the assert with `1` and use `IPloneSiteRoot.__identifier__` in not. The line with `alsoProvides` is no longer needed.

wesleybl: I think we could use `plone.app.contenttypes.interfaces.IEvent` in place of `IContentish`, and keep using `INavigationRoot` in the `not`. So the `alsoProvides` would be in `self.training` instead of `self.portal.events`.

jaroel:

jaroel:
https://jenkins.plone.org/job/pull-request-6.1-3.11/336/ is running this PR + https://github.com/plone/Products.CMFPlone/pull/3836

wesleybl:
@jaroel thanks for taking care of this!

Now we no longer have IContentish and we no longer need to run the jobs from this PR together with the CMFPlone PR. I started job 6.1-3.11 with just this PR.

news/518.bugfix
```diff
@@ -0,0 +1 @@
+Update test assertions for object_provides index due to Plone Site changes. @Akshat2Jain
```
wesleybl, line 1: @jaroel Put your id here too.

src/plone/api/tests/test_content.py
```diff
@@ -209,8 +208,8 @@ def test_create_dexterity(self):
         """Test create dexterity."""
         container = self.portal
 
-        # This section check for DX compatibilty. The custom DX types defined
-        # in plone.api are for Plone 4 compatiblity.
+        # This section check for DX compatibility. The custom DX types defined
+        # in plone.api are for Plone 4 compatibility.
```
wesleybl, line 212: I know it's not related to this PR but seeing "Plone 4" here made me see that this test and custom type can be removed. This branch is no longer compatible with Plone 4.

This can be done in another PR of course.

wesleybl (changes requested):

src/plone/api/tests/test_content.py
jaroel: @wesleybl Feel free to do so after landing this.

wesleybl:

jaroel:

wesleybl (approved):
LGTM

wesleybl:
@Akshat2Jain @davisagli @jaroel thank!

PR #519, opened 2023-10-14
Update GHA

gforcada:
- remove Plone 5.2
- remove Python 3.7
- add Python 3.10 and 3.11
- update the actions versions to avoid a deprecation warning

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -10,11 +10,11 @@ jobs:
 
     steps:
       # git checkout
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       # python setup
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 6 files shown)


gforcada:
As I'm only changing GHA and tox, we don't really need to run the Jenkins jobs.

I'm merging it myself to unlock the other PRs that are currently failing 🍀

mauritsvanrees:
Thanks.

But declaring 5.2 unsupported, even though no actual code was changed in this PR, is a breaking change: I think up until now the master branch could be used on 5.2 if you use Python 3.  This is fine to change, and there is no problem to include this in Plone 6.0.

But can you add a breaking news snippet and bump the version?

gforcada:
Sorry, I missed this for so many months 😖 is it still worth adding now the breaking change news entry, or _the ship already sailed_? 😓

PR #522, opened 2023-11-07
Issue 312: unrestricted find

gogobd:
Maybe we want this; unrestricted find is necessary because proxy roles are not being handled correctly atm.


docs/content.md
```diff
@@ -195,6 +195,21 @@ documents = api.content.find(
 %
 % self.assertGreater(len(documents), 0)
 
+Find all Documents, and use unrestrictedSearchResults:
```
stevepiercy, line 198:
```suggestion
Find all `Document` content types, using unrestricted search results:
```

docs/content.md
```diff
@@ -195,6 +195,21 @@ documents = api.content.find(
...
+
+```python
+from plone import api
+documents = api.content.find(
+    context=api.portal.get(),
+    portal_type='Document',
```
stevepiercy, line 204: Blackify

```suggestion
    portal_type="Document",
```

news/312.implementation
```diff
@@ -0,0 +1 @@
+Implemented unrestricted find.
```
stevepiercy, line 1: Give yourself credit!

```suggestion
Implemented unrestricted find of content types. @gogobd
```

stevepiercy (changes requested):
I reviewed docs only. A maintainer should review further.

src/plone/api/tests/test_content.py
```diff
@@ -912,6 +912,13 @@ def test_find(self):
         documents = api.content.find(portal_type="Document")
         self.assertEqual(len(documents), 2)
 
+    def test_find(self):
```
ale-rt, line 915: Hi @gogobd, good idea :) 
This test masks the one defined in line 908. It should have another name. In addition, it would be good to verify we find a different number of results in case we ask for an unrestricted search :)

src/plone/api/content.py
```diff
@@ -652,4 +652,7 @@ def find(context=None, depth=None, **kwargs):
     if not valid_indexes:
         return []
 
-    return catalog(**query)
+    if kwargs.get("unrestricted"):
```
ale-rt, line 655: I would update the function signature to have `unrestricted=False` as a default, otherwise you will be passing `unrestricted` to the catalog search.

ale-rt (changes requested):

gogobd:
In https://github.com/plone/plone.api/pull/523 @ksuess suggested to stick to one Pull Request, so I reopen...

gogobd:

news/312.implementation
davisagli, line 1: @gogobd This file's extension needs to match one of the predefined categories from https://github.com/plone/plone.api/blob/master/pyproject.toml

Use `.feature` for this one.

davisagli (dismissed):
@gogobd Thanks, this looks like a helpful improvement. Just needs one small fix to the name of the towncrier file.

src/plone/api/tests/test_content.py
mauritsvanrees: The function was renamed, which is good.  Well, I see a typo in the name: `test_unTrestricted_find`.

But it still finds the same number of results with or without `unrestricted=True`, so it does not tell us much.  Can you add a document that is not found in `test_find`, but is found in `test_unrestricted_find`?  I guess set an expiry date.

gogobd: @mauritsvanrees An expiry date in the past did't do it; I don't know how to create an object that is only found with "unrestricted=True"...

gogobd: I tried .start and .end and .effective and .expires; none of those made a difference...

davisagli: @gogobd The catalog's searchResults method does not filter by effective/expires if the current user has the "Access inactive portal content" permission (see https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/CatalogTool.py#L405), so it won't work if your test is running with the Manager role.

I'd suggest:
1. create a document that you do not publish
2. use setRoles to switch current roles to Anonymous
3. now the document should be missing with unrestricted=False, because it's filtered out based on the allowedRolesAndUsers index (https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/CatalogTool.py#L403)

gogobd: I just logged out and that did the trick!

gogobd:

gogobd:
I have no idea what this "tox linkcheck" wants to do, please help.

stevepiercy:
@gogobd see https://github.com/plone/plone.api/blob/master/tox.ini#L186-L197

stevepiercy:
Argh, it's *that* thing again. I'll push a commit to this PR that will fix it.

```
  The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

stevepiercy:
Actually, I'll push a separate PR so it has its own changelog, and you can cherry pick it.

stevepiercy:
Forget that, it already was fixed two weeks ago in https://github.com/plone/plone.api/commit/4b91ad80f4d96a92b374f7c3e12de7ae65fa5628, which was released in plone.api `2.0.9`.

Hmm, for some reason `tox -e linkcheck` does not use the pinned version of `sphinxcontrib-applehelp==1.0.4` in https://github.com/plone/plone.api/blob/master/requirements-docs.txt#L14C1-L14C14, but instead uses `sphinxcontrib-applehelp==1.0.8`. Why is that?

https://github.com/plone/plone.api/actions/runs/7981002081/job/21791817987?pr=522#step:6:14 says `linkcheck: install_deps> python -I -m pip install -r requirements-docs.txt`.

Later that gets clobbered in https://github.com/plone/plone.api/actions/runs/7981002081/job/21791817987?pr=522#step:6:23 with `linkcheck: install_package> python -I -m pip install --force-reinstall --no-deps /home/runner/work/plone.api/plone.api/.tox/.tmp/package/1/plone.api-2.0.7.dev0-0.editable-py3-none-any.whl`. **Why is it `2.0.7` and not `2.0.9`?**

This is above my pay grade for how to resolve.

wesleybl:
The branch needs to be updated.

stevepiercy (commented):
Docs pass now. LGTM.

ale-rt (approved):

mauritsvanrees (approved):
Thanks!

gogobd:
YAY! Thanks, every 01!

Issue #372, opened 2017-05-23
plone.api.user.has_permission lacks sanity check


jaroel:
> (Pdb) from plone import api
> (Pdb) api.user.has_permission('kek')
> True
> (Pdb) api.user.has_permission('dadddadas')
> True

I would expect an exception as those permission do not exists. One could argue that False should be returned for all non-existing permissions, but I'd rather have an error because my code no longer is correct.

Any thoughts?

adamcheasley:
Yes, that certainly confusing! I would go for an exception.

adamcheasley:
Looks like at the moment this function is just a wrapper around a portal_membership method:
```        
return bool(portal_membership.checkPermission(permission, obj))
```

jaroel:
Yep. Somewhere is a short circuit for a Manager to return True.
For normal users checkPermission does return False.

Issue #401, opened 2018-03-28
make api.user.has_permission work with permission ids too


frisi:
currently `api.user.has_permission('cmf.ModifyPortalContent')` returns `True` if logged in as `Manager`  and `False` for users having the `Modify portal content` permission.

problem: the method can only handle permission titles and silently fails when called with permission ids (https://github.com/zopefoundation/Products.CMFCore/blob/2.3.0/Products/CMFCore/permissions.zcml#L30)

i previously used the following code in my project:

```python
from AccessControl.security import checkPermission

def allowed(context):
    return checkPermission('cmf.ModifyPortalContent', context)
```

when i started to replace this checks with `plone.api.user.has_permission` i did not notice there is a problem first because i was logged in as manager user.
now our customer found out that ui elements where missing for content editors because the updated permission checks did not work properly.

this is why i'd suggest to make plone.api.user.has_permission to work with permission titles as well as ids
if that is not possible it would also help when `plone.api.user.has_permission('invalid permission string', context)` raises an exception if the permission is not valid (i can file another ticket for this if you agree with that)

some pseudo-test-code:

```python
setRoles(['Contributor'])
assertEqual(
   api.user.has_permission('cmf.ModifyPortalContent'),
   api.user.has_permission('Modify portal content'),
assertRaises(
  api.user.has_permission,
  'non existent permission',
  PermissionNotFoundError
) 
```

`plone.api.user.has_permission('invalid permission string', context)` should raise an exception if the permission id is not valid (i can file another ticket if you agree with that)

frisi:
i also assign @jensens because he recently contributed a change for has_permission (https://github.com/plone/plone.api/commit/c950845f07fea19d4b755cbd6085c2e62298ac5a)

jensens:
I agree we should support both writings. It is confusing enough to have those two different ways to do the same. Also, checking if the permission exists at all is a very good feature and would make problems with typos in permissions visible early.

jaroel:
Also see #372

mauritsvanrees:
👍 for supporting both, and for raising a `ValueError` or a more specific error in case no such permission is found.
I would say the same exception should be raised when this check is done for a Manager.
As anyone up for making a PR?

PR #526, opened 2023-12-22
Report non existing permissions

### LINKED ISSUE #515: Warn if the permission does not exist
I was doing some tests using `api.user.has_permission` and the test was failing 💥 but of course I should be right 🤣 

Turns out that I have a permission:

```xml
  <permission
    id="der.freitag.my-permission"
    title="der.freitag: Cool stuff"
    />
```

And my api call was:

```python
api.user.has_permission(
    'der.freitag.my-permission',
    obj=self.portal.autoren[user1['USER_ID']])
)
```

But, actually this is indeed wrong, the right call is:

```python
api.user.has_permission(
    'der.freitag: Cool stuff',
    obj=self.portal.autoren[user1['USER_ID']])
)
```

i.e. one has to use the `title` of the permission, not the `id`! 🤦🏾 

It would be nice if either `plone.api` could use either of them, or otherwise warn the user if the permission does not exist at all.

I can hardly think of any valid/reasonable use case where you don't want the permission that you are checking to __not__ exist.

As a first step, mentioning it on the documentation it would be helpful ✨

gforcada:
Closes #515

### DIFF
--- a/news/515.feature
+++ b/news/515.feature
@@ -1,0 +1,3 @@
+Report if a permission does not exist
+when calling `api.user.has_permission`.
+[gforcada]
--- a/src/plone/api/tests/test_env.py
+++ b/src/plone/api/tests/test_env.py
@@ -530,14 +530,15 @@ class TestPloneApiEnv(unittest.TestCase):
 
     def test_roles_restored_after_exception(self):
         """Tests that roles are restored after an exception."""
-        self.assertFalse(api.user.has_permission("Manage portal content"))
+        permission = "Manage properties"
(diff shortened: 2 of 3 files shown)


davisagli:
There's a failing test.

gforcada:
This is using #512 to be able to test properly with `tox`, so let's merge that one first 🍀

gforcada:

davisagli (approved):

ale-rt (approved):
I like the idea!

jaroel:
Closes #372

jaroel:
Addresses #401

PR #527, opened 2024-01-14
ankuristic/issues_492

Ankuristic:
Userid is not in UUID when set site to use UUID for userid.

To change the uuid for user id

### DIFF
--- a/src/plone/api/user.py
+++ b/src/plone/api/user.py
@@ -17,6 +17,7 @@ from Products.PlonePAS.interfaces.plugins import ILocalRolesPlugin
 
 import random
 import string
+import uuid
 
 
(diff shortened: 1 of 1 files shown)


Ankuristic:
Userid is not in UUID when set site to use UUID for userid

i only change in the plone.api.user there i used uuid with user_id

stevepiercy:
@Ankuristic this PR is harmful to Plone because it obliterates expected default behavior. No one asked for this. Closing.

Also please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

PR #528, opened 2024-01-31
Issue 479: Edited api.content.get_view for increased efficiency to get views

### LINKED ISSUE #479: Speed up api.content.get_view
The function does some avoidable checks before trying to wake up the view.
Those checks should be done only in case of error.
https://github.com/plone/plone.api/blob/3183dac2e84bb231a16f2da601a816aa80da7dd4/src/plone/api/content.py#L513-L536

samriddhi99:
The function does some avoidable check to wake up the view. I have used a try-except block avoid doing the check unless there is an error that pops up.
FIxes #479

### DIFF
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -523,20 +523,19 @@ def get_view(name=None, context=None, request=None):
     # available, because the __init__ of said view will contain
     # errors in client code.
 
-    # Get all available views...
-    sm = getSiteManager()
-    available_views = sm.adapters.lookupAll(
-        required=(providedBy(context), providedBy(request)),
-        provided=Interface,
-    )
-    # and get their names.
-    available_view_names = [view[0] for view in available_views]
-
-
     try:
-        return getMultiAdapter((context, request), name=name)
-    except Exception:  
-    # Check if the requested view is available.
+        adapter = getMultiAdapter((context, request), name=name)
+    except Exception: 
+
+        #Getting all available views
+        sm = getSiteManager()
+        available_views = sm.adapters.lookupAll(
+            required=(providedBy(context), providedBy(request)),
+            provided=Interface,
+        )
+        
+        # Check if the requested view is available by getting the names of all available views
+        available_view_names = [view[0] for view in available_views]
         if name not in available_view_names:
             # Raise an error if the requested view is not available.
             raise InvalidParameterError(
@@ -547,8 +546,7 @@ def get_view(name=None, context=None, request=None):
                     views="\n".join(sorted(available_view_names)),
                 ),
             )
-    return getMultiAdapter((context, request), name=name)
-
+    return adapter
 
 @required_parameters("obj")
 def get_uuid(obj=None):
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -532,16 +532,21 @@ def get_view(name=None, context=None, request=None):
     # and get their names.
     available_view_names = [view[0] for view in available_views]
 
-    # Raise an error if the requested view is not available.
-    if name not in available_view_names:
-        raise InvalidParameterError(
-            "Cannot find a view with name '{name}'.\n"
-            "Available views are:\n"
-            "{views}".format(
-                name=name,
-                views="\n".join(sorted(available_view_names)),
-            ),
-        )
+
+    try:
+        return getMultiAdapter((context, request), name=name)
+    except Exception:  
+    # Check if the requested view is available.
+        if name not in available_view_names:
+            # Raise an error if the requested view is not available.
+            raise InvalidParameterError(
+                "Cannot find a view with name '{name}'.\n"
+                "Available views are:\n"
+                "{views}".format(
+                    name=name,
+                    views="\n".join(sorted(available_view_names)),
+                ),
+            )
     return getMultiAdapter((context, request), name=name)
 
 
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -535,7 +535,8 @@ def get_view(name=None, context=None, request=None):
             provided=Interface,
         )
         
-        # Check if the requested view is available by getting the names of all available views
+        # Check if the requested view is available
+        # by getting the names of all available views
         available_view_names = [view[0] for view in available_views]
         if name not in available_view_names:
             # Raise an error if the requested view is not available.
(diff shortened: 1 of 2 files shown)


samriddhi99:

src/plone/api/content.py
```diff
@@ -532,16 +532,21 @@ def get_view(name=None, context=None, request=None):
...
+            # Raise an error if the requested view is not available.
+            raise InvalidParameterError(
+                "Cannot find a view with name '{name}'.\n"
+                "Available views are:\n"
+                "{views}".format(
+                    name=name,
+                    views="\n".join(sorted(available_view_names)),
+                ),
+            )
     return getMultiAdapter((context, request), name=name)
```
davisagli, line 550: If the view is found, we end up getting the adapter twice. It should be assigned to a variable above so that it can simply be returned here.

src/plone/api/content.py
```diff
@@ -532,16 +532,21 @@ def get_view(name=None, context=None, request=None):
...
-    if name not in available_view_names:
-        raise InvalidParameterError(
-            "Cannot find a view with name '{name}'.\n"
-            "Available views are:\n"
-            "{views}".format(
-                name=name,
-                views="\n".join(sorted(available_view_names)),
-            ),
-        )
+
```
davisagli, line 535: The above logic to get available_view_names should also only be done if it is needed.

davisagli (changes requested):

src/plone/api/content.py
samriddhi99: Thank you for your feedback. I have made some changes accordingly. Hope it is alright.

samriddhi99:

gforcada:
@samriddhi99 thanks for this contribution, hopefully we get all the ticks green ✅ and can merge it soon! 🤩 

There is one tick that you need to work on though: **the contributor agreement**, please read about it and sign it 🤞🏾 https://plone.org/foundation/contributors-agreement

src/plone/api/content.py
```diff
@@ -523,27 +523,30 @@ def get_view(name=None, context=None, request=None):
...
+            # Raise an error if the requested view is not available.
+            raise InvalidParameterError(
+                "Cannot find a view with name '{name}'.\n"
+                "Available views are:\n"
+                "{views}".format(
+                    name=name,
+                    views="\n".join(sorted(available_view_names)),
+                ),
+            )
+    return adapter
```
gforcada, line 549: `adapter` here could be not defined, if we get an exception when trying to get the adapter, and thus get into the `exception Exception` part of the code and the view name it is found in the `available_view_names`.

Could that happen, though? 🤔 A view that you don't have permissions for would get you in such corner case? 🤔

stevepiercy:
@gforcada it appears @samriddhi99 was approved and added to the Contributors team after mister-roboto sent the reminder.

stevepiercy:
Something is not quite right. @samriddhi99 was added to the Plone Contributors Team in GitHub, but mister-roboto says otherwise. It's possible that the emails in @samriddhi99's GitHub account do not align with what they put on their Agreement.

@samriddhi99 would you please add the email you put on your agreement to your GitHub account? If you already have it, it is possible that there is a typo somewhere, and you should contact agreements@plone.org, copying and pasting a link to this issue in reference to this issue.

src/plone/api/content.py
```diff
@@ -523,27 +523,30 @@ def get_view(name=None, context=None, request=None):
...
-        raise InvalidParameterError(
-            "Cannot find a view with name '{name}'.\n"
-            "Available views are:\n"
-            "{views}".format(
-                name=name,
-                views="\n".join(sorted(available_view_names)),
-            ),
+    try:
+        adapter = getMultiAdapter((context, request), name=name)
+    except Exception:
```
ale-rt, line 528: I think this should be `except ComponentLookupError`, otherwise you risk to return a wrong message.

src/plone/api/content.py
ale-rt: The security is not checked when you call `getMultiAdapter`, anyway the `except Exception` is not good IMO.

src/plone/api/content.py
samriddhi99: Thank you for pointing it out! I have made the changes accordingly.

src/plone/api/content.py
```diff
@@ -523,27 +525,29 @@ def get_view(name=None, context=None, request=None):
...
-    if name not in available_view_names:
-        raise InvalidParameterError(
-            "Cannot find a view with name '{name}'.\n"
-            "Available views are:\n"
-            "{views}".format(
-                name=name,
-                views="\n".join(sorted(available_view_names)),
-            ),
+    try:
+        adapter = getMultiAdapter((context, request), name=name)
```
ale-rt, line 529:
```suggestion
        return getMultiAdapter((context, request), name=name)
```

src/plone/api/content.py
```diff
@@ -523,27 +525,29 @@ def get_view(name=None, context=None, request=None):
...
+            # Raise an error if the requested view is not available.
+            raise InvalidParameterError(
+                "Cannot find a view with name '{name}'.\n"
+                "Available views are:\n"
+                "{views}".format(
+                    name=name,
+                    views="\n".join(sorted(available_view_names)),
+                ),
+            )
+    return adapter
```
ale-rt, line 550:
```suggestion
```

ale-rt (commented):
LGTM, I added a couple of suggestions to have micro-optimizations.
Thanks for your contribution!

samriddhi99:
@stevepiercy 
The plone contributors agreement verifier check has passed, so I think the issue has been resolved.

stevepiercy:
@samriddhi99 thank you, and I confirm. You're not the only person that experienced a similar issue.

After all other checks pass, please read and follow this comment: https://github.com/plone/plone.api/pull/528#issuecomment-1918427786

samriddhi99:

src/plone/api/content.py
```diff
@@ -523,27 +525,28 @@ def get_view(name=None, context=None, request=None):
...
+        #Getting all available views
+        sm = getSiteManager()
+        available_views = sm.adapters.lookupAll(
+            required=(providedBy(context), providedBy(request)),
+            provided=Interface,
         )
-    return getMultiAdapter((context, request), name=name)
-
+        
+        # Check if the requested view is available by getting the names of all available views
```
ale-rt, line 538: We tend to follow the black line length limit of 88 characters:
```suggestion
        # Check if the requested view is available
        # by getting the names of all available views
```
Usually we have `pre-commit` to check this, but it is not yet in this package.
So do this at will

ale-rt (approved):

davisagli (approved):
Thanks for this contribution!

PR #535, opened 2024-04-29
In relation.create, handle existing value of None

### LINKED ISSUE #534: `plone.api.relation.create` error on not required RelationList field
**Steps to reproduce**

- Provide a behavior or extend a schema with a not required `RelationList` field:

```
    some_related_stuff = RelationList(
        title="Some not required related stuff",
        value_type=RelationChoice(vocabulary='plone.app.vocabularies.Catalog'),
        required=False,
    )
```
- It is important not to set `default=[]`, which according to the [documentation](https://5.docs.plone.org/develop/plone/forms/schemas.html#field-constructor-parameters) is a dangerous default. ([IRelatedItems](https://github.com/plone/plone.app.relationfield/blob/master/plone/app/relationfield/behavior.py#L18) has it 10 years now though?!?)
- Create an object `obj1` with having no `some_related_stuff` relations, aka `None` value.
- Try to create a new relation using `plone.api.relation.create`.

```
api.relation.create(source=obj1, target=obj2, relationship='some_related_stuff')
```

**What happens**

It goes into an error, since it gets `None` here:

https://github.com/plone/plone.api/blob/08884b755450cdc5290d640b826d3f659025179f/src/plone/api/relation.py#L220

**What I expect**

A relation created.

davisagli:
Fixes #534

### DIFF
--- a/news/535.bugfix
+++ b/news/535.bugfix
@@ -1,0 +1,1 @@
+In relation.create: Fix edge case where existing RelationList value is None. @davisagli
--- a/src/plone/api/relation.py
+++ b/src/plone/api/relation.py
@@ -217,7 +217,7 @@ def create(source=None, target=None, relationship=None):
             target.absolute_url(),
         )
         if not has_relation:
-            existing_relations = getattr(source, from_attribute, [])
+            existing_relations = getattr(source, from_attribute, None) or []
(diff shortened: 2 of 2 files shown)


davisagli:

ale-rt (approved):

PR #539, opened 2024-07-20
Overhaul contributing documentation for Plone 6

### LINKED ISSUE #357: Update contribution docs
Currently, the docs on how to contribute still reference the makefile, which no longer exists. These should be updated so that others can easily contribute. I believe the suggested way to get a dev environment these days is to use the coredev buildout, is that correct?

### LINKED ISSUE #473: Replace Netlify preview build command with 'tox -e docs' and remove Makefile
Do you want to keep the Makefile or use tox?

_Originally posted by @stevepiercy in https://github.com/plone/plone.api/pull/469#discussion_r844825454_

'tox -e docs' needs tox to be installed.

stevepiercy:
Now that we have Plone core contributing documentation, I've simplified and modernized the contributing documentation.

See https://github.com/plone/documentation/issues/1278.

- Closes #473
- Closes #357

After merging, then merge https://github.com/plone/documentation/pull/1684.


stevepiercy:

stevepiercy:
I have no clue as to why Jenkins fails for Plone 6.0/Python 3.11 but not for Plone 6.0/Python 3.8. Help!

docs/contribute.md
```diff
@@ -0,0 +1,244 @@
...
+Alternatively, you can automatically reload changes to the documentation as you edit it in a web browser.
+
+```shell
+tox -e livehtml
+```
+
+The [`plone.api` documentation](https://6.docs.plone.org/plone.api) is automatically generated from the documentation source files when its submodule is updated in the [main Plone `documentation` repository](https://github.com/plone/documentation/).
+
+
+### Add a function to an existing module
```
ksuess, line 134: This is a 2nd level heading, not a subheading under "Documentation"

ksuess, line 134:
```suggestion
### Add a method to an existing module
```

ksuess (changes requested):
I appreciate your overhaul. Just minor suggestions to content that was already there before this PR.

docs/contribute.md
```diff
@@ -0,0 +1,244 @@
...
+## Continuous integration
+
+`plone.api` uses GitHub workflows for continuous integration.
+On every push to the `master` branch, GitHub runs its workflows for all tests and code quality checks.
+GitHub workflows are configured in the directory `.github/workflows` at the root of this package.
+
+
+## Documentation
+
+For every feature change or addition to `plone.api`, you should add documentation of it.
```
jensens, line 112: `you should add documentation`-> it's a `must` here, isn't  it?

docs/contribute.md
jensens: Well, the whole plone.api is build with functions (not classes+methods), so it is explained below. So IMO functions is fine.

jensens (commented):
thanks! Minor things see comments.

docs/contribute.md
stevepiercy: Fixed in b4e53e8c081d08906b3e5092017d72a424f2c991.

stevepiercy:

stevepiercy:
Changes made, and I excluded GNU make as a requirement here since this project uses tox instead.

It would be nice to use one of make or tox for all of Plone, don't you think?

ksuess:
> It would be nice to use one of make or tox for all of Plone, don't you think?

Yes, a switch to `make` would be consistent. Definitly.

docs/contribute.md
ksuess: The implementation of a feature is of course a function, not a classes method. That's why I wrote this header like this back then.
It's just that I realised that in the rest of the documentation it is often spoken of 'API methods'.
But I am fine with keeping the header like it is.

ksuess (approved):

stevepiercy:
> > It would be nice to use one of make or tox for all of Plone, don't you think?
> 
> Yes, a switch to `make` would be consistent. Definitly.

See https://community.plone.org/t/tools-for-development-gnu-make-tox-or/19708

PR #550, opened 2024-12-20
Fix api.content.get when a item in the path is not acccessible

### LINKED ISSUE #549: api.content.get no longer returns content that is inside a inaccessible container
Given a private container `/Plone/cannot-view` and a published item `/Plone/cannot-view/can-view` I cannot traverse to the item any more since Plone 5.2:

```python
with api.env.adopt_roles(["Member"]):
    api.content.get(path="/Plone/cannot-view/can-view")
```

```python
with api.env.adopt_roles(["Member"]):
    portal.restrictedTraverse("/Plone/cannot-view/can-view")
```

I think that breaks user-expectations of the api and should be changed. 

In Plone 5.1 it still works, since Plone 5.2 it raises Unauthorized. The reason may be that a different `guarded_getattr` seems to be used in `restrictedTraverse` but I'm not sure - the import is the same but the executed code is different. 

In a project I instead used unrestricedTraverse and then checked the View-permission for the object:

```python
obj = portal.unrestrictedTraverse(path)
if obj and api.user.has_permission("View", obj=obj):
    ...
```

pbauer:
Fixes #549

----
📚 Documentation preview 📚: https://ploneapi--550.org.readthedocs.build/

### DIFF
--- a/news/549.bugfix
+++ b/news/549.bugfix
@@ -1,0 +1,2 @@
+Fix api.content.get(path=path) when a item in the path is not accessible to the user.
+[pbauer]
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -133,7 +133,12 @@ def get(path=None, UID=None):
                 relative_path=path,
             )
         try:
-            content = site.restrictedTraverse(path)
+            path = path.split("/")
+            if len(path) > 1:
+                parent = site.unrestrictedTraverse(path[:-1])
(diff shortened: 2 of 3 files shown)


pbauer:

ale-rt (approved):
To be sure we do not have regression a test would be nice, but this looks good

pbauer:

PR #551, opened 2024-12-20
 backport fix for #549

### LINKED ISSUE #435: set_registry_record WrongType handling
**What steps reproduce the problem?**
* Create a Registry Schema with a field not named "field_one".
* Write some function that uses "set_registry_record" from the api and set an other type than the defined for the field

**What do you expect when you do that?**
I get an Exception of type InvalidParameterError

**What happens instead?**
Exception of type IndexError "Index out of range"

**Which Plone version are you using?**
5.2

**Cause of the Bug**
https://github.com/plone/plone.api/blob/8da902ea95abc56cf4923146f7e38b360562f864/src/plone/api/portal.py#L378

**Potential fix**
`if field[0] == name`

### LINKED ISSUE #549: api.content.get no longer returns content that is inside a inaccessible container

pbauer:
Fixes #549 for Plone 5.2

### DIFF
--- a/news/551.bugfix
+++ b/news/551.bugfix
@@ -1,0 +1,2 @@
+Backport fix for #549
+[pbauer]
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -157,7 +157,12 @@ def get(path=None, UID=None):
             )
 
         try:
-            return site.restrictedTraverse(path)
+            path = path.split('/')
+            if len(path) > 1:
+                parent = site.unrestrictedTraverse(path[:-1])
+                return parent.restrictedTraverse(path[-1])
+            else:
+                return site.restrictedTraverse(path[-1])
         except (KeyError, AttributeError):
             return None  # When no object is found don't raise an error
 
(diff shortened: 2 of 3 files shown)


pbauer:

mauritsvanrees:
I have released 1.11.2 with this.

PR #570, opened 2025-03-09
Fixes random temporary id creation leading to `zExceptions.BadRequest` error

### LINKED ISSUE #445: content.create: Random id generation is unsafe
/src/plone/api/content.py, lines 81/82:

    # Create a temporary id if the id is not given
    content_id = not safe_id and id or str(random.randint(0, 99999999))

We have a website with thousands of articles which are imported as Plone objects every couple of weeks and have numeric names which we pass as title parameter to api.content.create. Sometimes the random temporary id will conflict with an id of a previously generated object, resulting in a BadRequest error.

So instead of blindly using the random temporary id a test should be done if this id can be used safely, by checking if an object with this id already exists in the container. How about this:

```
<   content_id = not safe_id and id or str(random.randint(0, 99999999))
>   while (True):
>        content_id = not safe_id and id or str(random.randint(0, 99999999))
>        if content_id not in container.keys():
>            break
```

This solution definately isn't perfect, but if you have items with most or all possible ids from 0 to 99999999 in one folder, you will have plenty of other issues anyway...

rohnsha0:
fixes #445 

----
📚 Documentation preview 📚: https://ploneapi--570.org.readthedocs.build/


src/plone/api/content.py
```diff
@@ -67,7 +70,18 @@ def create(
...
     # Create a temporary id if the id is not given
-    content_id = not safe_id and id or str(random.randint(0, 99999999))
+    if not safe_id and id:
+        content_id = id
+    else:
+        # Try to generate a unique random ID
+        attempts = 0
+        while attempts < MAX_UNIQUE_ID_ATTEMPTS:
+            content_id = str(random.randint(0, 99999999))
+            if content_id not in container.keys():
```
ale-rt, line 80:
```suggestion
            if content_id not in container:
```

ale-rt (approved):
LGTM, just a minor suggestion

src/plone/api/content.py
rohnsha0: DONE!

rohnsha0:

src/plone/api/content.py
```diff
@@ -67,7 +70,18 @@ def create(
...
+    else:
+        # Try to generate a unique random ID
+        attempts = 0
+        while attempts < MAX_UNIQUE_ID_ATTEMPTS:
+            content_id = str(random.randint(0, 99999999))
+            if content_id not in container:
+                break
+            attempts += 1
+        # If we couldn't find a unique ID after max attempts, use the last one anyway
+        # This maintains the original behavior if all attempts fail
```
ale-rt, line 84: Here we could even decide to raise an exception, I will leave this decision to other reviewers

ale-rt (approved):

src/plone/api/content.py
```diff
@@ -67,7 +70,18 @@ def create(
     :Example: :ref:`content-create-example`
     """
     # Create a temporary id if the id is not given
-    content_id = not safe_id and id or str(random.randint(0, 99999999))
+    if not safe_id and id:
```
stevepiercy, line 73: Is there a reason to use an int instead of https://docs.python.org/3/library/uuid.html#uuid.uuid4 from the Python standard library? I don't know the underlying data type.

See the introduction from https://docs.python.org/3/library/uuid.html

> If all you want is a unique ID, you should probably call [uuid1()](https://docs.python.org/3/library/uuid.html#uuid.uuid1) or [uuid4()](https://docs.python.org/3/library/uuid.html#uuid.uuid4). Note that [uuid1()](https://docs.python.org/3/library/uuid.html#uuid.uuid1) may compromise privacy since it creates a UUID containing the computer’s network address. [uuid4()](https://docs.python.org/3/library/uuid.html#uuid.uuid4) creates a random UUID.

news/445.bugfix
```diff
@@ -0,0 +1 @@
+fix random temporary id creation leading to `zExceptions.BadRequest` error @rohnsha0
```
stevepiercy, line 1: This may change, subsequent to review from more knowledgeable Plonistas than I.

```suggestion
Attempt to generate a random temporary id for a content type up to 100 times, else continue to raise a `zExceptions.BadRequest` error. @rohnsha0
```

stevepiercy (changes requested):

src/plone/api/content.py
rohnsha0: I kept the underlying logic same...

mauritsvanrees: Since we are touching this, I would also be in favour of `str(uuid.uuid4())`.

src/plone/api/content.py
mauritsvanrees: I would indeed say it is better to raise an explicit error here, instead of encountering a less understandable error later on.  According to issue #445 you would get a BadRequest error.  But there is nothing wrong with the request, so we should use something else.  Maybe:

```
raise ValueError("Could not find unique id while creating content."
```

src/plone/api/content.py
ale-rt: I would not go for a 32 characters id.

If you really need to use a UUID, I propose:
```
>>> uuid.uuid4().hex[:8]
'bba42d98'
```

stevepiercy: > >>> uuid.uuid4().hex[:8]
> 'bba42d98'

Is that guaranteed to be unique? The goal would be to remove the while loop to ensure uniqueness.

ale-rt: > > > > > uuid.uuid4().hex[:8]
> > > > > 'bba42d98'
> 
> Is that guaranteed to be unique? The goal would be to remove the while loop to ensure uniqueness.

No, it can never be guaranteed to be unique, not even if you use the 32 character long one.
But we have 16^8 values to pick from, which is something like 2 ^ 32 which is more that 10 ^ 9.

rohnsha0: > > > > > > uuid.uuid4().hex[:8]
> > > > > > 'bba42d98'
> > 
> > Is that guaranteed to be unique? The goal would be to remove the while loop to ensure uniqueness.
> 
> No, it can never be guaranteed to be unique, not even if you use the 32 character long one.
> But we have 16^8 values to pick from, which is something like 2 ^ 32 which is more that 10 ^ 9.

Which in my opinion will be enough!

src/plone/api/content.py
rohnsha0: Makes sense!

src/plone/api/content.py
stevepiercy: Keep in mind that you're evaluating against millions of objects, not just one. See the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_problem).

I also don't like playing odds, and prefer certainty. [uuid4()](https://docs.python.org/3/library/uuid.html#uuid.uuid4) is unique.

What's unmanageable about it? You click or tap the URL, and you're done.

mauritsvanrees: Remember that the id is temporary.  A few lines further on, we use the title to generate a nicer id.  So in my opinion the temporary id can be as ugly as we want.  So let's use the full uuid4.  And sure, retry 100 times if needed.

rohnsha0: done

stevepiercy (approved):
Thank you!

stevepiercy:
I'm not sure why Jenkins didn't run when @rohnsha0 requested it. Maybe permissions are not aligned? Let's see if my username can jumpstart Jenkins.

stevepiercy:

stevepiercy:
For @rohnsha0's run, it looks like Jenkins actually ran and passed, but didn't return results back to GitHub. Can someone take a look and verify?

https://jenkins.plone.org/job/pull-request-6.0-3.9/121/robot/

rohnsha0:
> I'm not sure why Jenkins didn't run when @rohnsha0 requested it. Maybe permissions are not aligned? Let's see if my username can jumpstart Jenkins.

Ah, that ran and it was green too! But I merged the main branch into this branch. It's counted as a commit, so all tests needed to run again! 😅

rohnsha0:
the jenkins is bleeding yet again. out of nowhere!

src/plone/api/content.py
ale-rt: > Remember that the id is temporary. A few lines further on, we use the title to generate a nicer id. So in my opinion the temporary id can be as ugly as we want. So let's use the full uuid4. And sure, retry 100 times if needed.

I am pretty sure that I saw some content type (probably some that did not require a title), for which the generated id was kept.
BTW, I do not want to make this an argument. We anyway need to change the logic to [avoid moving the object after its creation](https://github.com/plone/plone.api/issues/439).
But that is happening in another PR.

mauritsvanrees (approved):
LGTM now, thanks!

Those Jenkins runs were having trouble with the robot browser tests again.  Let me try again:

mauritsvanrees:
Again a robot testing problem on Jenkins node 3 (which is my node).  Has nothing to do with this PR.  I merge.  Thanks!

PR #592, opened 2025-11-30
Add typing annotations

ale-rt:
Initial draft courtesy of `monkeytype`:

```shell
.venv/bin/uv pip install -e plone.api[test] monkeytype zope.testrunner
monkeytype run .venv/bin/zope-testrunner --path plone.api/src/
for module in `monkeytype list-modules`; do monkeytype apply $module || true > /dev/null;done
```

- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't any other open pull requests for the same change.
- [x] I followed the guidelines in [Contributing to Plone](https://6.docs.plone.org/contributing/index.html).
- [x] I successfully ran code quality checks on my changes locally.
- [x] I successfully ran tests on my changes locally.
- [x] If needed, I added new tests for my changes.
- [x] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

If your pull request closes an open issue, include the exact text below, immediately followed by the issue number. When your pull request gets merged, then that issue will close automatically.

Closes #

----
📚 Documentation preview 📚: https://ploneapi--592.org.readthedocs.build/

### DIFF
--- a/.meta.toml
+++ b/.meta.toml
@@ -9,6 +9,15 @@ commit-id = "2.9.0"
 codespell_extra_lines = """
         exclude: docs/locale/.*.pot
 """
+extra_lines = """
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.18.2
+    hooks:
+    -   id: mypy
+        additional_dependencies:
+            - types-decorator
+        exclude: docs
+"""
 
 [pyproject]
 check_manifest_ignores = """
@@ -23,7 +32,7 @@ check_manifest_ignores = """
     "requirements-docs.txt",
     "requirements.txt",
 """
-dependencies_ignores = "['Products.PrintingMailHost', 'plone.app.iterate',]"
+dependencies_ignores = "['Products.DCWorkflow', 'Products.PrintingMailHost', 'plone.app.iterate',]"
 codespell_ignores = "manuel,checkin"
 
 [tox]
--- a/src/plone/api/addon.py
+++ b/src/plone/api/addon.py
@@ -11,6 +11,7 @@ from plone.base.interfaces import INonInstallable
 from plone.base.utils import get_installer
 from Products.CMFPlone.controlpanel.browser.quickinstaller import InstallerView
 from Products.GenericSetup import EXTENSION
+from typing import Any
 from zope.component import getAllUtilitiesRegisteredFor
 from zope.globalrequest import getRequest
 
@@ -94,7 +95,7 @@ def _get_non_installable_addons() -> NonInstallableAddons:
 
 
 @lru_cache(maxsize=1)
-def _cached_addons() -> tuple[tuple[str, AddonInformation]]:
+def _cached_addons() -> tuple[tuple[str, AddonInformation], ...]:
     """Return information about add-ons in this installation.
 
     :returns: Tuple of tuples with add-on id and AddonInformation.
@@ -130,7 +131,7 @@ def _cached_addons() -> tuple[tuple[str, AddonInformation]]:
         profile_type = pid_parts[-1]
         if product_id not in addons:
             # get some basic information on the product
-            product = {
+            product: dict[str, Any] = {
                 "id": product_id,
                 "version": get_version(product_id),
                 "title": product_id,
@@ -271,10 +272,10 @@ def get(addon: str) -> AddonInformation:
     :returns: Add-on information.
     :rtype: string
     """
-    addons = dict(_cached_addons())
+    addons: dict[str, AddonInformation] = dict(_cached_addons())
     if addon not in addons:
         raise InvalidParameterError(f"No add-on {addon} found.")
-    return _update_addon_info(addons.get(addon), _get_installer())
+    return _update_addon_info(addons[addon], _get_installer())
 
 
 @required_parameters("addon")
--- a/src/plone/api/content.py
+++ b/src/plone/api/content.py
@@ -2,10 +2,15 @@
 
 from Acquisition import aq_chain
 from Acquisition import aq_inner
+from collections.abc import Callable
+from collections.abc import Iterator
 from copy import copy as _copy
 from itertools import islice
 from plone.api import portal
 from plone.api.exc import InvalidParameterError
+from plone.api.types import Container
+from plone.api.types import Content
+from plone.api.types import Request
 from plone.api.validation import at_least_one_of
 from plone.api.validation import mutually_exclusive_parameters
 from plone.api.validation import required_parameters
@@ -14,6 +19,9 @@ from plone.app.uuid.utils import uuidToObject
 from plone.uuid.interfaces import IUUID
 from Products.CMFCore.DynamicType import DynamicType
 from Products.CMFCore.WorkflowCore import WorkflowException
+from Products.CMFPlone.WorkflowTool import WorkflowTool
+from Products.DCWorkflow.DCWorkflow import DCWorkflowDefinition
+from typing import Any
 from zope.component import ComponentLookupError
 from zope.component import getMultiAdapter
 from zope.component import getSiteManager
@@ -21,6 +29,10 @@ from zope.container.interfaces import INameChooser
 from zope.globalrequest import getRequest
 from zope.interface import Interface
 from zope.interface import providedBy
+from zope.interface.interface import InterfaceClass
+from ZPublisher.BaseRequest import RequestContainer
+from ZTUtils.Lazy import LazyCat
+from ZTUtils.Lazy import LazyMap
 
 import transaction
 import uuid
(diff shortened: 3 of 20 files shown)


ale-rt:

davisagli (commented):
This will take some time to review carefully.

For a start, at first glance, I see lots of annotations with ImplicitAcquisitionWrapper that look wrong. Those are usually meant to be a DexterityContent, and I think monkeytype was confused by the acquisition wrapper.

ale-rt:
> Most of these APIs are designed specifically to work with content objects, not any acquisition-wrapped Python object. So using ImplicitAcquisitionWrapper still seems quite wrong to me.

To me it seems wrong to say it returns a dexterity content :D
I could replace that with `Any`, because that would be accurate. What do you think?

ale-rt:

news/+ec620103.internal.md
davisagli, line 1: Remove this duplicate changelog entry

src/plone/api/_types.py
```diff
@@ -0,0 +1,20 @@
+"""Shared internal typing aliases used across plone.api modules."""
+
+from Acquisition import ImplicitAcquisitionWrapper
+from typing import TypeAlias
+from ZPublisher.HTTPRequest import HTTPRequest
+
+try:
+    from plone.dexterity.content import DexterityContent as _DexterityContext
+except ImportError:  # pragma: no cover
+    _DexterityContext = ImplicitAcquisitionWrapper
```
davisagli: plone.dexterity is a hard dependency of plone.api in setup.py. So the conditional import isn't needed.

src/plone/api/_types.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+# Public-facing context type for content APIs.
+Context: TypeAlias = _DexterityContext
```
davisagli: I would call this Content instead of Context. There are plenty of contexts in Zope which are not content.

src/plone/api/_types.py
```diff
@@ -0,0 +1,20 @@
...
+
+# Broader acquisition-wrapped context for APIs that can accept portal/tool roots.
+ContainerContext: TypeAlias = ImplicitAcquisitionWrapper
```
davisagli: I need to check where this is used, but maybe it should be Item or ObjectManager to follow Zope naming.

src/plone/api/addon.py
```diff
@@ -274,7 +276,9 @@ def get(addon: str) -> AddonInformation:
     addons = dict(_cached_addons())
     if addon not in addons:
         raise InvalidParameterError(f"No add-on {addon} found.")
-    return _update_addon_info(addons.get(addon), _get_installer())
+    return _update_addon_info(
+        cast(AddonInformation, addons.get(addon)), _get_installer()
```
davisagli: Would it work to declare the type of `addons` as `dict[str, AddonInformation]` above?

src/plone/api/content.py
```diff
@@ -34,13 +45,13 @@
 @required_parameters("container", "type")
 @at_least_one_of("id", "title")
 def create(
-    container=None,
-    type=None,
-    id=None,
-    title=None,
-    safe_id=False,
+    container: ContainerContext | None = None,
```
davisagli: I think this should be Products.CMFCore.PortalFolder.PortalFolder. That is where the invokeFactory method is defined.

src/plone/api/content.py
```diff
@@ -34,13 +45,13 @@
...
+    type: str | None = None,
+    id: str | None = None,
```
davisagli, line 51: Why are type and container optional when we have the `required_paremeters` decorator saying they are required?

The repeated use of `cast(ContainerContext, container)` below feels like an ugly workaround for the fact that the type includes None when that is not really possible.

src/plone/api/_types.py
davisagli: Also I wonder if we should use Products.CMFCore.PortalContent.PortalContent here instead of DexterityContent

src/plone/api/content.py
```diff
@@ -165,11 +183,17 @@ def get(path=None, UID=None):
...
     elif UID:
         return uuidToObject(UID)
+    return None
 
 
 @required_parameters("source")
 @at_least_one_of("target", "id")
-def move(source=None, target=None, id=None, safe_id=False):
+def move(
+    source: Context | None = None,
```
davisagli: Same question about why this has a default None when it is required.

src/plone/api/content.py
```diff
@@ -327,7 +367,7 @@ def delete(obj=None, objects=None, check_linkintegrity=True):
 
 
 @required_parameters("obj")
-def get_state(obj=None, default=_marker):
+def get_state(obj: Context | None = None, default: Any = _marker) -> str:
```
davisagli: I was thinking default should be a str, but I guess we also need to handle code that passes a marker object. We could define our own marker type and ask callers to use an instance of that so we can type this as `str | Marker`

src/plone/api/content.py
```diff
@@ -419,7 +466,12 @@ def _wf_transitions_for(workflow, from_state, to_state):
     return _find_path(transition_maps, [], to_state, from_state)
 
 
-def _transition_to(obj, workflow, to_state, **kwargs):
+def _transition_to(
+    obj: Context,
+    workflow: Any,
```
davisagli: workflow has to be WorkflowTool from CMFCore

src/plone/api/portal.py
```diff
@@ -51,7 +60,7 @@
 MISSING = object()
 
 
-def get():
+def get() -> ContainerContext:
```
davisagli: Should be PloneSite

src/plone/api/_types.py
ale-rt: Good point, I think that given that we have an alias it is pretty easy to change this at any time

src/plone/api/addon.py
ale-rt: Yes, thanks! This is way better.

src/plone/api/content.py
ale-rt: yes, that should be solved now

.meta.toml
```diff
@@ -9,6 +9,15 @@ commit-id = "2.5.1"
 codespell_extra_lines = """
         exclude: docs/locale/.*.pot
 """
+YesYesyyyyyyyy = """
```
davisagli: ???

src/plone/api/content.py
```diff
@@ -34,13 +46,13 @@
 @required_parameters("container", "type")
 @at_least_one_of("id", "title")
 def create(
-    container=None,
-    type=None,
-    id=None,
-    title=None,
-    safe_id=False,
+    container: Content,
```
davisagli: I think we should use PortalFolder here. We use the invokeFactory method, which comes from there.

.meta.toml
ale-rt: LOL, what's that 🤣

davisagli:

davisagli (approved):
LGTM, assuming the jenkins builds pass

ale-rt:
@ericof, @mauritsvanrees are you ok having this merged?

ericof:
Would you mind taking a look at plone-stubs? It solves many issues I see with your current approach.
Either way, I'm not opposed to your merge

davisagli:
@ericof I worked on this with @ale-rt at the sprint. We resolved the issues I had noticed earlier and I think it's in pretty good shape now. What are the issues you still see? We can't consider fixing them if you don't say what they are.

You and I have talked about plone-stubs before. I see the reasoning for having one package in the short term where stubs can be added for multiple Plone packages. But I also think in the long term we should try to add type hints to the actual packages.

ericof:
@davisagli, @ale-rt First, sorry for sounding so "negative", that was not my intention.

Instead of doing a nit-pick about which types should be returned, I would like to suggest two immediate points (considering what I've learned so far about typing `plone.api` and other packages) to be addressed:

- We probably should be using more `@overloads` to closely match the signature of the functions here:

```python
@overload
def get(userid: str, username: str | None = None) -> MemberData | None:
@overload
def get(userid: None, username: str) -> MemberData | None:
```

- To avoid polluting the code with all overloads, we should ship the  `.pyi` files

If `plone.api` is completely typed I would remove it from plone-stubs and keep working on the other packages.

And, as I said, I'm not opposing this PR

davisagli:
@ericof Thanks for the clarification. I agree, using overloads for the functions that accept several different combinations of parameters sounds like it would be a nice improvement.

davisagli:
Let's go ahead and merge this, and work on other improvements (such as overloads) in a new PR.

